PR #7, opened 2018-08-20
Added Opensource config

hiranya911:


.opensource/project.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "Firebase Admin SDK - .NET",
+  "platforms": [
+      ".NET",
```
samtstern: I've got a feeling this will cause a parse error ... how about `DotNet` or `CSharp`?  Something alphanumeric

hiranya911: Done

PR #10, opened 2018-10-25
Reduce lock contention on HttpPublicKeySource.GetPublicKeysAsync

cmpcdaly:
Implemented a double-checked lock so that the semaphore handle isn't obtained unless the token is null.

### DIFF
--- a/FirebaseAdmin/FirebaseAdmin.Tests/Auth/FirebaseAuthTest.cs
+++ b/FirebaseAdmin/FirebaseAdmin.Tests/Auth/FirebaseAuthTest.cs
@@ -136,7 +136,7 @@ namespace FirebaseAdmin.Auth.Tests
             var canceller = new CancellationTokenSource();
             canceller.Cancel();
             var idToken = await FirebaseTokenVerifierTest.CreateTestTokenAsync();
-            await Assert.ThrowsAsync<OperationCanceledException>(
+            await Assert.ThrowsAnyAsync<OperationCanceledException>(
                 async () => await FirebaseAuth.DefaultInstance.VerifyIdTokenAsync(
                     idToken, canceller.Token));
         }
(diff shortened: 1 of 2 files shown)


cmpcdaly:
I signed it!

FirebaseAdmin/FirebaseAdmin/Auth/HttpPublicKeySource.cs
```diff
@@ -66,36 +66,41 @@ public HttpPublicKeySource(string certUrl, IClock clock, HttpClientFactory clien
...
-                        response.EnsureSuccessStatusCode();
-                        _cachedKeys = ParseKeys(await response.Content.ReadAsStringAsync()
-                            .ConfigureAwait(false));
-                        var cacheControl = response.Headers.CacheControl;
-                        if (cacheControl != null && cacheControl.MaxAge.HasValue)
+                        using (var httpClient = _clientFactory.CreateDefaultHttpClient())
                         {
-                            _expirationTime = now.Add(cacheControl.MaxAge.Value)
-                                .Subtract(ClockSkew);
+                            var now = _clock.UtcNow;
```
hiranya911: Lets move this out of the `if` block (just inside the `try` block):

```
var now = _clock.UtcNow;
if (_cachedKeys == null || now >= _expirationTime)
{

}
```

FirebaseAdmin/FirebaseAdmin.Tests/Auth/FirebaseAuthTest.cs
```diff
@@ -136,7 +136,7 @@ public async Task VerifyIdTokenCancel()
             var canceller = new CancellationTokenSource();
             canceller.Cancel();
             var idToken = await FirebaseTokenVerifierTest.CreateTestTokenAsync();
-            await Assert.ThrowsAsync<OperationCanceledException>(
+            await Assert.ThrowsAnyAsync<OperationCanceledException>(
```
hiranya911: Why was this change necessary? Can we revert it?

cmpcdaly: As we're now passing the CancellationToken to the Semaphore, this will throw a TaskCanceledException if the given CancellationToken has been cancelled (which is a more derived type of OperationCanceledException https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskcanceledexception)

cmpcdaly:
Thanks for the feedback, I've made the requested changed to HttpPublicKeySource

PR #12, opened 2018-12-06
FCM Implementation (Part 1)

hiranya911:
go/dotnet-fcm

Implements the following parts of the API:

```
FireabseMessaging
    SendAsync()
Message
Notification
AndroidConfig
ApnsConfig
WebpushConfig
```

Rest of the API will be fleshed out in subsequent PRs.

PR #13, opened 2018-12-13
Implemented AndroidNotification API

hiranya911:
Follow up to #12


hiranya911:
@stewartmiles I've documented all internal types and properties (some of it was done in #12, and merged back here).

PR #14, opened 2018-12-19
Some clean up and refactoring of existing FCM code

hiranya911:
1. Renamed `Message.AndroidConfig` property to `Message.Android` (to be consistent with the proposal go/dotnet-fcm).
2. Removed the internal `Validated*` types, and absorbed the corresponding serialization/deserialization logic into the existing public types.
3. Implemented stronger copy-and-validate semantics for messages. Each message property is copied prior to validation. This ensures that any mutations done by user code to the original `Message` or its properties, won't affect the json posted to the FCM service.

### DIFF
--- a/FirebaseAdmin/FirebaseAdmin/Messaging/AndroidConfig.cs
+++ b/FirebaseAdmin/FirebaseAdmin/Messaging/AndroidConfig.cs
@@ -28,54 +28,21 @@ namespace FirebaseAdmin.Messaging
         /// messages that can be collapsed, so that only the last message gets sent when delivery can be
         /// resumed. A maximum of 4 different collapse keys may be active at any given time.
         /// </summary>
+        [JsonProperty("collapse_key")]
         public string CollapseKey { get; set; }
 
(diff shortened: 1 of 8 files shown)


FirebaseAdmin/FirebaseAdmin/Messaging/AndroidConfig.cs
```diff
@@ -89,26 +56,41 @@ private string PriorityString
...
+            set
+            {
+                switch (value)
+                {
+                    case "high":
+                        Priority = Messaging.Priority.High;
+                        return;
+                    case "normal":
+                        Priority = Messaging.Priority.High;
+                        return;
```
bklimt: Seems like there should be some error handling here for other values?

stewartmiles: +1

hiranya911: Done

PR #16, opened 2019-01-10
Implemented FCM Webpush  API

hiranya911:
* Implemented `WebpushConfig` and `WebpushNotification` types
* Removed an unused extension method (`IEnumerable.Copy()`)
* Made some internal properties private for better encapsulation

### DIFF
--- a/FirebaseAdmin/FirebaseAdmin/Messaging/WebpushNotification.cs
+++ b/FirebaseAdmin/FirebaseAdmin/Messaging/WebpushNotification.cs
@@ -154,18 +154,18 @@ namespace FirebaseAdmin.Messaging
         public int[] Vibrate { get; set; }
 
         /// <summary>
+        /// A collection of notification actions to be associated with the notification.
+        /// </summary>
+        [JsonProperty("actions")]
+        public IEnumerable<Action> Actions;
+
+        /// <summary>
         /// A collection of arbitrary key-value data to be included in the notification. This is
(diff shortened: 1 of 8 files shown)


FirebaseAdmin/FirebaseAdmin/Messaging/WebpushNotification.cs
```diff
@@ -0,0 +1,276 @@
...
+        /// Any arbitrary data that should be associated with the notification.
+        /// </summary>
+        [JsonProperty("data")]
+        public object Data { get; set; }
+
+        /// <summary>
+        /// The direction in which to display the notification.
+        /// </summary>
+        [JsonIgnore]
+        public Direction? Direction { get; set; }
```
stewartmiles: I'm wondering why you have a set of nullable properties for types like this?  Why not use default values instead?  To me it seems like a pain to deal with the nullability of e.g bool properties rather than just reading the default value.

hiranya911: This is to make sure those parameters get dropped from the JSON output. For most parameters, we want to avoid sending a default value to the backend (0 for ints, false for bools etc) when they are not set by the user.

stewartmiles: It does worry me a little that we're letting optimizations leak into the API surface.  I guess if that's the way you do it across the board it's ok to me.

PR #19, opened 2019-01-23
Implemented APNs Support in the API

hiranya911:
* Implemented the types ApnsConfig, Aps, ApsAlert and CriticalSound


FirebaseAdmin/FirebaseAdmin/Messaging/CriticalSound.cs
```diff
@@ -0,0 +1,91 @@
...
+        /// <summary>
+        /// Integer representation of the <see cref="Critical"/> property, which is how
+        /// APNs expects it.
+        /// </summary>
+        [JsonProperty("critical")]
+        private int? CriticalInt
+        {
+            get
+            {
+                if (Critical)
```
stewartmiles: If Critical is false should this return 0 instead?

PR #34, opened 2019-03-20
Proxy Support

nixj14:
Added support for configuring HttpClientFactory.  Needed for Proxy support.  Also provide further configuration capabilities now that HttpClientFactory is exposed.

### DIFF
--- a/FirebaseAdmin/FirebaseAdmin.Tests/FirebaseAppTest.cs
+++ b/FirebaseAdmin/FirebaseAdmin.Tests/FirebaseAppTest.cs
@@ -104,7 +104,7 @@ namespace FirebaseAdmin.Tests
             Assert.Same(credential, copy.Credential);
             Assert.Equal("test-project", copy.ProjectId);
             Assert.Equal("test@service.account", copy.ServiceAccountId);
-            Assert.Equal(typeof(MockHttpClientFactory), copy.HttpClientFactory.GetType());
+            Assert.Same(typeof(MockHttpClientFactory), copy.HttpClientFactory.GetType());
         }
 
         [Fact]
@@ -125,7 +125,21 @@ namespace FirebaseAdmin.Tests
             Assert.Same(credential, copy.Credential);
             Assert.Equal("test-project", copy.ProjectId);
             Assert.Equal("test@service.account", copy.ServiceAccountId);
-            Assert.Equal(typeof(HttpClientFactory), copy.HttpClientFactory.GetType());
+            Assert.Same(typeof(HttpClientFactory), copy.HttpClientFactory.GetType());
+        }
+
+        [Fact]
+        public void NoHttpClientFactory()
+        {
+            var credential = GoogleCredential.FromAccessToken("token");
+            var options = new AppOptions()
+            {
+                Credential = credential,
+                ProjectId = "test-project",
+                ServiceAccountId = "test@service.account",
+                HttpClientFactory = null,
+            };
+            Assert.Throws<ArgumentNullException>(() => FirebaseApp.Create(options));
         }
 
         [Fact]
--- a/FirebaseAdmin/FirebaseAdmin.Tests/FirebaseAppTest.cs
+++ b/FirebaseAdmin/FirebaseAdmin.Tests/FirebaseAppTest.cs
@@ -16,6 +16,7 @@ using System;
 using System.Threading.Tasks;
 using FirebaseAdmin;
 using Google.Apis.Auth.OAuth2;
+using Google.Apis.Http;
 using Xunit;
 
 namespace FirebaseAdmin.Tests
@@ -93,6 +94,7 @@ namespace FirebaseAdmin.Tests
                 Credential = credential,
                 ProjectId = "test-project",
                 ServiceAccountId = "test@service.account",
+                HttpClientFactory = new MockHttpClientFactory(new MockMessageHandler()),
             };
             var app = FirebaseApp.Create(options);
             Assert.Equal("[DEFAULT]", app.Name);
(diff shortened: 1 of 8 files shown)


nixj14:
I signed it!

nixj14:
I signed it!

FirebaseAdmin/FirebaseAdmin/AppOptions.cs
```diff
@@ -58,5 +60,10 @@ internal AppOptions(AppOptions options)
         /// JSON.</para>
         /// </summary>
         public string ServiceAccountId { get; set; }
+
+        /// <summary>
+        /// Gets or sets the HttpClientFactory to use when making Firebase requirests.
+        /// </summary>
+        public HttpClientFactory ClientFactory { get; set; } = new HttpClientFactory();
```
hiranya911: Lets call it `HttpClientFactory` without shortening the name.

hiranya911: As I understand this is C# 6 syntax. But we do target some .NET versions that are before C# 6. I'm not sure if that's a problem. The safe solution here would be to assign the default value in the constructor. WDYT?

nixj14: I'm happy to change, though I did notice that FirebaseUserManagerArgs uses "ClientFactory" as well.  In the end, I'm indifferent, but would recommend sticking with one naming convention or the other.

nixj14: Making the changes now.

hiranya911: We should be consistent with other Google libraries like Google API client, at least in our public API surface: https://github.com/googleapis/google-api-dotnet-client/blob/master/Src/Support/Google.Apis/Services/BaseClientService.cs#L68

+1 to renaming the internal fields too.

FirebaseAdmin/FirebaseAdmin/AppOptions.cs
```diff
@@ -25,6 +26,8 @@ namespace FirebaseAdmin
     /// </summary>
     public sealed class AppOptions
     {
+        private HttpClientFactory httpClientFactory;
```
hiranya911: Can we do this instead?

```
public AppOptions()
{
  this.HttpClientFactory = new HttpClientFactory();
}

public HttpClientFactory HttpClientFactory { get; set; }
```

Then in `FirebaseApp` constructor:

```
if (this.options.HttpClientFactory == null)
{
  // throw
}
```

Without an explicit throw, we could accidentally mask an error in user code.

FirebaseAdmin/FirebaseAdmin.Tests/FirebaseAppTest.cs
```diff
@@ -102,6 +104,28 @@ public void CreateAppOptions()
             Assert.Same(credential, copy.Credential);
             Assert.Equal("test-project", copy.ProjectId);
             Assert.Equal("test@service.account", copy.ServiceAccountId);
+            Assert.Equal(typeof(MockHttpClientFactory), copy.HttpClientFactory.GetType());
```
hiranya911: We can do a stronger `Assert.Same()` check here.

FirebaseAdmin/FirebaseAdmin.Tests/FirebaseAppTest.cs
```diff
@@ -102,6 +104,28 @@ public void CreateAppOptions()
...
+            };
+            var app = FirebaseApp.Create(options);
+            Assert.Equal("[DEFAULT]", app.Name);
+
+            var copy = app.Options;
+            Assert.NotSame(options, copy);
+            Assert.Same(credential, copy.Credential);
+            Assert.Equal("test-project", copy.ProjectId);
+            Assert.Equal("test@service.account", copy.ServiceAccountId);
+            Assert.Equal(typeof(HttpClientFactory), copy.HttpClientFactory.GetType());
```
hiranya911: Perhaps a not null check is enough here?

Also add a test case to verify that FirebaseApp doesn't allow null values for HttpClientFactory.

FirebaseAdmin/FirebaseAdmin.Tests/Messaging/FirebaseMessagingTest.cs
```diff
@@ -54,6 +54,17 @@ public void GetMessaging()
...
 
+        [Fact]
+        public void GetMessagingWithClientFactory()
+        {
+            var app = FirebaseApp.Create(new AppOptions() { Credential = MockCredential, HttpClientFactory = new HttpClientFactory() }, "MyApp");
+            FirebaseMessaging messaging = FirebaseMessaging.GetMessaging(app);
+            Assert.NotNull(messaging);
+            Assert.Same(messaging, FirebaseMessaging.GetMessaging(app));
+            app.Delete();
+            Assert.Throws<InvalidOperationException>(() => FirebaseMessaging.GetMessaging(app));
```
hiranya911: We should try to do a stronger check here. Otherwise a future code change can silently break this functionality. Easiest option would be to inject a `MockHttpClientFactory` via `AppOptions`. Then invoke `FirebaseMessaging.SendAsync()` and see if that factory get called.

FirebaseAdmin/FirebaseAdmin.Tests/FirebaseAppTest.cs
```diff
@@ -104,7 +104,7 @@ public void CreateAppOptions()
             Assert.Same(credential, copy.Credential);
             Assert.Equal("test-project", copy.ProjectId);
             Assert.Equal("test@service.account", copy.ServiceAccountId);
-            Assert.Equal(typeof(MockHttpClientFactory), copy.HttpClientFactory.GetType());
+            Assert.Same(typeof(MockHttpClientFactory), copy.HttpClientFactory.GetType());
```
hiranya911: `Assert.Same(options.HttpClientFactory, copy.HttpClientFactory);`

hiranya911:
Sorry for the delay. Approval for this API just came through. I will merge it shortly.

PR #36, opened 2019-03-21
feature: multicast messaging

kentcb:
Add support for multicast/send-all messaging.


FirebaseAdmin/FirebaseAdmin.IntegrationTests/FirebaseMessagingTest.cs
```diff
@@ -49,5 +49,76 @@ public async Task Send()
...
+                },
+                Tokens = new[]
+                {
+                    "token1",
+                    "token2",
+                },
+            };
+            var response = await FirebaseMessaging.DefaultInstance.SendMulticastAsync(multicastMessage, dryRun: true);
+            Assert.NotNull(response);
+            Assert.Equal(2, response.FailureCount);
```
kentcb: Can't actually test success case because tokens are required, and there doesn't seem to be any support for "dummy" testing tokens.

FirebaseAdmin/FirebaseAdmin/Messaging/FirebaseMessagingException.cs
```diff
@@ -0,0 +1,41 @@
...
+        internal FirebaseMessagingException(int errorCode, string message, Exception inner)
+            : base(message, inner)
+        {
+            this.ErrorCode = errorCode;
+        }
+
+        /// <summary>
+        /// Gets an error code that may provide more information about the error.
+        /// </summary>
+        public int ErrorCode { get; }
```
kentcb: I'm not sure I've done the right thing here. The Java code seems to expose the error code as a string rather than int, but it gets it by mapping some other string that I don't seem to have access to (or don't know where to get it).

kentcb:
@hiranya911 Ready for review.

FirebaseAdmin/FirebaseAdmin/Messaging/FirebaseMessagingException.cs
hiranya911: There's some on going work to revamp/redesign how errors are exposed in our SDKs. In the meantime our plan is to not introduce new exception types to the public API. So I think we can just remove this class, and use FirebaseException with a message string instead.

FirebaseAdmin/FirebaseAdmin.IntegrationTests/FirebaseMessagingTest.cs
hiranya911: Right. FCM does not support this at the moment. There's some ongoing work in the backend to enable this. In the meantime, checking the error case is enough. Since `SendMulticast` uses `SendAll` under the hood, we are not losing much anyway.

hiranya911:
Thank you @kentcb. Love what you've done here. I will take a closer look tomorrow morning. If all goes well I think we can release this next week (this API was already proposed and approved at the same time I worked on the Node and Java implementations).

FirebaseAdmin/FirebaseAdmin.Tests/Messaging/MulticastMessageTest.cs
```diff
@@ -0,0 +1,1542 @@
...
+    public class MulticastMessageTest
+    {
+        [Fact]
+        public void EmptyMulticastMessage()
+        {
+            var message = new MulticastMessage
+            {
+                Tokens = new[] { "test-token" },
+            };
+            this.AssertJsonEquals(new JObject() { { "tokens", new JArray("test-token") } }, message);
```
hiranya911: This object never really gets json serialized directly. We should be unit testing the `GetMessageList()` method here.

FirebaseAdmin/FirebaseAdmin.Tests/Messaging/MulticastMessageTest.cs
```diff
@@ -0,0 +1,1542 @@
...
+            var message = new MulticastMessage
+            {
+                Tokens = new[] { "test-token" },
+                Data = new Dictionary<string, string>()
+                {
+                    { "k1", "v1" },
+                    { "k2", "v2" },
+                },
+            };
+            this.AssertJsonEquals(
```
hiranya911: Same comment as above. Unit test GetMessageList and make sure we correctly create 2 copies of the message. Applies to most test cases in this class.

FirebaseAdmin/FirebaseAdmin.Tests/Messaging/FirebaseMessagingClientTest.cs
```diff
@@ -89,6 +88,65 @@ public async Task SendAsync()
             Assert.Equal(2, handler.Calls);
         }
 
+        [Fact]
+        public async Task SendAllAsync()
```
hiranya911: Also add unit tests for null list, empty list and list longer than 100. And a mock response that contains some errors.

FirebaseAdmin/FirebaseAdmin/Messaging/FirebaseMessagingClient.cs
```diff
@@ -129,5 +210,75 @@ internal class SendResponse
...
+            private readonly bool dryRun;
+
+            public FCMClientServiceRequest(IClientService clientService, string restPath, Message message, bool dryRun)
+                : base(clientService)
+            {
+                this.restPath = restPath;
+                this.message = message;
+                this.dryRun = dryRun;
+
+                this.InitParameters();
```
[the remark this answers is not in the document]
kentcb: Alas, no. The `ClientServiceRequest` assumes it is non-`null`, and the only means of ensuring it is non-`null` is to invoke `InitParameters`.

FirebaseAdmin/FirebaseAdmin/Messaging/MulticastMessage.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// serialized into the JSON format expected by the FCM service. Each property is copied
+        /// before validation to guard against the original being modified in the user code
+        /// post-validation.
+        /// </summary>
+        internal MulticastMessage CopyAndValidate()
+        {
+            // Copy and validate the leaf-level properties
+            var copy = new MulticastMessage
+            {
+                Tokens = this.Tokens?.Copy(),
```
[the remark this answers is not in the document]
kentcb: Hmm, that won't deal with `null`, so I did this:
```C#
new List<string>(this.Tokens ?? Enumerable.Empty<string>())
```

FirebaseAdmin/FirebaseAdmin/Messaging/MulticastMessage.cs
```diff
@@ -0,0 +1,122 @@
...
+    /// Contains payload information as well as the list of device registration tokens to which the
+    /// message should be sent. A single <c>MulticastMessage</c> may contain up to 100 registration
+    /// tokens.
+    /// </summary>
+    public sealed class MulticastMessage
+    {
+        /// <summary>
+        /// Gets or sets the registration tokens for the devices to which the message should be distributed.
+        /// </summary>
+        [JsonProperty("tokens")]
```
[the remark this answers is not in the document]
kentcb: Oh wow, yeah. This helped me realise that the vast majority of the tests for `MulticastMessage` were actually testing `Message` functionality, so I've removed them.

FirebaseAdmin/FirebaseAdmin.Tests/Messaging/MulticastMessageTest.cs
kentcb: Totally, per my comment above 🤦‍♂️

kentcb:
@hiranya911 Thanks very much for the review. I've pushed fixes based on your comments.

FirebaseAdmin/FirebaseAdmin/Messaging/SendResponse.cs
```diff
@@ -0,0 +1,57 @@
...
+using System.Linq;
+using Google.Apis.Util;
+
+namespace FirebaseAdmin.Messaging
+{
+    /// <summary>
+    /// Response from an operation that sends FCM messages to multiple recipients.
+    /// See <see cref="FirebaseMessaging.SendMulticastAsync(MulticastMessage)"/>.
+    /// </summary>
+    public sealed class SendResponse
```
hiranya911: This should be called BatchResponse.

FirebaseAdmin/FirebaseAdmin/Messaging/SendItemResponse.cs
```diff
@@ -0,0 +1,73 @@
...
+
+namespace FirebaseAdmin.Messaging
+{
+    /// <summary>
+    /// The result of an individual send operation that was executed as part of a batch. See
+    /// <see cref="SendResponse"/> for more details.
+    /// </summary>
+    public sealed class SendItemResponse
+    {
+        private SendItemResponse(string messageId)
```
hiranya911: This should be called SendResponse.

FirebaseAdmin/FirebaseAdmin/Messaging/FirebaseMessagingClient.cs
```diff
@@ -102,11 +112,97 @@ internal sealed class FirebaseMessagingClient : IDisposable
...
+        /// </summary>
+        /// <param name="messages">The messages to be sent. Must not be null.</param>
+        /// <param name="dryRun">A boolean indicating whether to perform a dry run (validation
+        /// only) of the send. If set to true, the messages will be sent to the FCM backend service,
+        /// but it will not be delivered to any actual recipients.</param>
+        /// <param name="cancellationToken">A cancellation token to monitor the asynchronous
+        /// operation.</param>
+        /// <returns>A task that completes with a <see cref="SendResponse"/>, giving details about
+        /// the batch operation.</returns>
+        public async Task<SendResponse> SendAllAsync(
```
hiranya911: The returned object should be called `BatchResponse` to be consistent with Java and Node.js. Sorry if I wasn't clear in my previous review. The public API should be:

```
class BatchResponse
{
  public int SuccessCount;
  public int FailureCount;
  public IReadOnlyList<SendResponse> Responses;
}

class SendResponse
{
  public string MessgeId;
  public FirebaseException Exception;
  public boolean IsSuccess;
}
```

FirebaseAdmin/FirebaseAdmin/Messaging/SendResponse.cs
```diff
@@ -0,0 +1,57 @@
...
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="SendResponse"/> class.
+        /// </summary>
+        /// <param name="responses">The responses.</param>
+        public SendResponse(IEnumerable<SendItemResponse> responses)
```
hiranya911: Lets make the constructor `internal`.

FirebaseAdmin/FirebaseAdmin/Messaging/MulticastMessage.cs
```diff
@@ -0,0 +1,99 @@
...
+        /// </summary>
+        public ApnsConfig Apns { get; set; }
+
+        internal List<Message> GetMessageList()
+        {
+            var tokens = this.Tokens;
+
+            if (tokens == null)
+            {
+                throw new InvalidOperationException("Tokens cannot be null.");
```
hiranya911: Lets check for both null and > 100, and throw an `ArgumentException`. That way if a developer passes in an invalid `MulticastMessage` to one of the send APIs, they will get an `ArgumentException`.

FirebaseAdmin/FirebaseAdmin/Messaging/FirebaseMessagingClient.cs
```diff
@@ -102,11 +112,97 @@ internal sealed class FirebaseMessagingClient : IDisposable
...
+            IEnumerable<Message> messages,
+            bool dryRun = false,
+            CancellationToken cancellationToken = default(CancellationToken))
+        {
+            messages.ThrowIfNull(nameof(messages));
```
hiranya911: This is duplicated now (see next line). Let's remove it.

FirebaseAdmin/FirebaseAdmin/Messaging/FirebaseMessagingClient.cs
```diff
@@ -121,13 +217,78 @@ internal class SendRequest
...
+
+            public override string HttpMethod => "POST";
+
+            public override string RestPath => this.restPath;
+
+            public override string MethodName => throw new NotImplementedException();
+
+            protected override object GetBody()
+            {
+                var sendRequest = new SendRequest
```
hiranya911: Nit: 
```
return new SendRequest()
{
  ...
};
```

FirebaseAdmin/FirebaseAdmin/Messaging/FirebaseMessaging.cs
```diff
@@ -167,6 +168,128 @@ public async Task<string> SendAsync(Message message, bool dryRun)
...
 
+        /// <summary>
+        /// Sends all the messages in the given list via Firebase Cloud Messaging. Employs batching to
+        /// send the entire list as a single RPC call. Compared to the <see cref="SendAsync(Message)"/>
+        /// method, this is a significantly more efficient way to send multiple messages.
+        /// </summary>
+        /// <param name="messages">Up to 100 messages to send in the batch. Cannot be null.</param>
+        /// <returns>A <see cref="SendResponse"/> containing details of the batch operation's
+        /// outcome.</returns>
+        public async Task<SendResponse> SendAllAsync(IEnumerable<Message> messages)
```
hiranya911: Renamed return type to BatchResponse.

kentcb:
@hiranya911 Ah, my bad. OK, I've fixed that as well as the other items you pointed out.

PR #43, opened 2019-04-01
Update UserRecord

levimatheri:
Please advise on necessary changes and/or tests


hiranya911:
Thanks @levimatheri. I would say, look at our Node.js or Java SDKs for some inspiration. We will probably need several different types. `UserInfo`, `UserRecord` and `ExportedUserRecord`. `UserInfo` could probably be an interface that `UserRecord` implements. `ExportedUserRecord` should extend `UserRecord`.

PR #46, opened 2019-04-08
Update user record

levimatheri:
@hiranya911 How does this look?


levimatheri:
@hiranya911 Travis has finally passed. Let me know what changes/improvements I should make. Some tests needed but I plan to add those as a separate PR

PR #49, opened 2019-05-01
Implemented FirebaseAuth.GetUser based on the Java implementation

renatoc8:
This pull request implements the GetUser() function in FirebaseAuth, closely following the implementation found in the firebase-admin-java project.

### DIFF
--- a/FirebaseAdmin/FirebaseAdmin/Auth/FirebaseAuth.cs
+++ b/FirebaseAdmin/FirebaseAdmin/Auth/FirebaseAuth.cs
@@ -267,14 +267,14 @@ namespace FirebaseAdmin.Auth
         /// Gets a <see cref="UserRecord"/> object containig information about the user who's
         /// user ID was specified in <paramref name="uid"/>.
         /// </summary>
-        /// <param name="uid">The user ID for the user who's data is to be retrieved..</param>
+        /// <param name="uid">The user ID for the user who's data is to be retrieved.</param>
         /// <returns>A task that completes with a <see cref="UserRecord"/> representing
         /// a user with the specified user ID.</returns>
         /// <exception cref="ArgumentException">If user ID argument is null or empty.</exception>
         /// <exception cref="FirebaseException">If a user cannot be found with the specified user ID.</exception>
         public async Task<UserRecord> GetUserAsync(string uid)
         {
-            return await this.userManager.Value.GetUserById(uid);
+            return await this.GetUserAsync(uid, default(CancellationToken));
         }
 
         /// <summary>
@@ -291,7 +291,9 @@ namespace FirebaseAdmin.Auth
         public async Task<UserRecord> GetUserAsync(
             string uid, CancellationToken cancellationToken)
         {
-            return await this.userManager.Value.GetUserById(uid, cancellationToken);
+            var userManager = this.IfNotDeleted(() => this.userManager.Value);
+
+            return await userManager.GetUserById(uid, cancellationToken);
         }
 
         /// <summary>
--- a/FirebaseAdmin/FirebaseAdmin/Auth/FirebaseAuth.cs
+++ b/FirebaseAdmin/FirebaseAdmin/Auth/FirebaseAuth.cs
@@ -264,6 +264,20 @@ namespace FirebaseAdmin.Auth
         }
 
         /// <summary>
+        /// Gets a <see cref="UserRecord"/> object containig information about the user who's
+        /// user ID was specified in <paramref name="uid"/>.
+        /// </summary>
+        /// <param name="uid">The user ID for the user who's data is to be retrieved..</param>
+        /// <returns>A task that completes with a <see cref="UserRecord"/> representing
+        /// a user with the specified user ID.</returns>
+        /// <exception cref="ArgumentException">If user ID argument is null or empty.</exception>
+        /// <exception cref="FirebaseException">If a user cannot be found with the specified user ID.</exception>
+        public async Task<UserRecord> GetUser(string uid)
+        {
+            return await this.userManager.Value.GetUserById(uid);
+        }
+
+        /// <summary>
         /// Sets the specified custom claims on an existing user account. A null claims value
         /// removes any claims currently set on the user account. The claims must serialize into
         /// a valid JSON string. The serialized claims must not be larger than 1000 characters.
--- a/FirebaseAdmin/FirebaseAdmin/Auth/FirebaseAuth.cs
+++ b/FirebaseAdmin/FirebaseAdmin/Auth/FirebaseAuth.cs
@@ -272,9 +272,26 @@ namespace FirebaseAdmin.Auth
         /// a user with the specified user ID.</returns>
         /// <exception cref="ArgumentException">If user ID argument is null or empty.</exception>
         /// <exception cref="FirebaseException">If a user cannot be found with the specified user ID.</exception>
-        public async Task<UserRecord> GetUser(string uid)
+        public async Task<UserRecord> GetUserAsync(string uid)
         {
             return await this.userManager.Value.GetUserById(uid);
+        }
+
+        /// <summary>
+        /// Gets a <see cref="UserRecord"/> object containig information about the user who's
+        /// user ID was specified in <paramref name="uid"/>.
+        /// </summary>
+        /// <param name="uid">The user ID for the user who's data is to be retrieved.</param>
+        /// <param name="cancellationToken">A cancellation token to monitor the asynchronous
+        /// operation.</param>
+        /// <returns>A task that completes with a <see cref="UserRecord"/> representing
+        /// a user with the specified user ID.</returns>
+        /// <exception cref="ArgumentException">If user ID argument is null or empty.</exception>
+        /// <exception cref="FirebaseException">If a user cannot be found with the specified user ID.</exception>
+        public async Task<UserRecord> GetUserAsync(
+            string uid, CancellationToken cancellationToken)
+        {
+            return await this.userManager.Value.GetUserById(uid, cancellationToken);
         }
 
         /// <summary>
--- a/FirebaseAdmin/FirebaseAdmin/Auth/IUserInfo.cs
+++ b/FirebaseAdmin/FirebaseAdmin/Auth/IUserInfo.cs
@@ -1,0 +1,50 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace FirebaseAdmin.Auth
+{
+    /// <summary>
+    /// A collection of standard profile information for a user. Used to expose profile information
+    /// returned by an identity provider.
+    /// </summary>
+    public interface IUserInfo
+    {
+        /// <summary>
+        /// Returns the user's unique ID assigned by the identity provider.
+        /// </summary>
+        /// <returns>a user ID string.</returns>
+        string GetUid();
+
+        /// <summary>
+        /// Returns the user's display name, if available.
+        /// </summary>
+        /// <returns>a display name string or null.</returns>
+        string GetDisplayName();
+
+        /// <summary>
+        /// Returns the user's email address, if available.
+        /// </summary>
+        /// <returns>an email address string or null.</returns>
+        string GetEmail();
+
+        /// <summary>
+        /// Returns the user's phone number, if available.
+        /// </summary>
+        /// <returns>a phone number string or null.</returns>
+        string GetPhoneNumber();
+
+        /// <summary>
+        /// Returns the user's photo URL, if available.
+        /// </summary>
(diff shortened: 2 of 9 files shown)


hiranya911:
Thanks @renatoc8. I think we'll need the exact API structure currently available in Java. Specifically, we'd want to have an `IUserInfo` interface and have `UserRecord` implement it. Providers should also implement the same interface.

Other high-level comments:
1. We can remove the `FirebaseAdmin.Auth.Internal` namespace. Just put those classes in the `Auth` namespace, and mark them as `internal`. 
2. The public method name should be `GetUserAsync()`. And we'd also need an overload that accepts a cancellation token.
3. All constructors of `UserRecord` should be internal.

renatoc8:
Thanks for reviewing my pull requests. I will get those changes implemented shortly.

renatoc8:
Ok, those changes have been implemented. Anything else you can think of @hiranya911 ?

FirebaseAdmin/FirebaseAdmin/Auth/FirebaseAuth.cs
```diff
@@ -263,6 +263,37 @@ public async Task<FirebaseToken> VerifyIdTokenAsync(string idToken)
...
+        /// user ID was specified in <paramref name="uid"/>.
+        /// </summary>
+        /// <param name="uid">The user ID for the user who's data is to be retrieved..</param>
+        /// <returns>A task that completes with a <see cref="UserRecord"/> representing
+        /// a user with the specified user ID.</returns>
+        /// <exception cref="ArgumentException">If user ID argument is null or empty.</exception>
+        /// <exception cref="FirebaseException">If a user cannot be found with the specified user ID.</exception>
+        public async Task<UserRecord> GetUserAsync(string uid)
+        {
+            return await this.userManager.Value.GetUserById(uid);
```
hiranya911: You need to wrap this in `IfNotDeleted()`. See `SetCustomUserClaimsAsync()` method for an example.

FirebaseAdmin/FirebaseAdmin/Auth/FirebaseAuth.cs
```diff
@@ -263,6 +263,37 @@ public async Task<FirebaseToken> VerifyIdTokenAsync(string idToken)
                 .ConfigureAwait(false);
         }
 
+        /// <summary>
+        /// Gets a <see cref="UserRecord"/> object containig information about the user who's
+        /// user ID was specified in <paramref name="uid"/>.
+        /// </summary>
+        /// <param name="uid">The user ID for the user who's data is to be retrieved..</param>
```
hiranya911: Remove extra period at the end of sentence.

FirebaseAdmin/FirebaseAdmin/Auth/IUserInfo.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// A collection of standard profile information for a user. Used to expose profile information
+    /// returned by an identity provider.
+    /// </summary>
+    public interface IUserInfo
+    {
+        /// <summary>
+        /// Returns the user's unique ID assigned by the identity provider.
+        /// </summary>
+        /// <returns>a user ID string.</returns>
+        string GetUid();
```
hiranya911: I think all these should be properties instead of methods.

FirebaseAdmin/FirebaseAdmin/Auth/ProviderUserInfo.cs
```diff
@@ -0,0 +1,71 @@
...
+            this.phoneNumber = provider.PhoneNumber;
+            this.photoUrl = provider.PhotoUrl;
+            this.providerId = provider.ProviderID;
+        }
+
+        /// <summary>
+        /// Returns the user's unique ID assigned by the identity provider.
+        /// </summary>
+        /// <returns>a user ID string.</returns>
+        public string GetUid() => this.uid;
```
hiranya911: Implement as properties.

FirebaseAdmin/FirebaseAdmin/Auth/UserMetadata.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// Contains additional metadata associated with a user account.
+    /// </summary>
+    public sealed class UserMetadata
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="UserMetadata"/> class with the specified creation and last sign-in timestamps.
+        /// </summary>
+        /// <param name="creationTimestamp">A timestamp representing the date and time that the user account was created.</param>
+        /// <param name="lastSignInTimestamp">A timestamp representing the date and time that the user account was last signed-on to.</param>
+        public UserMetadata(long creationTimestamp, long lastSignInTimestamp)
```
hiranya911: internal (at least for now; in the future we might have to expose this)

FirebaseAdmin/FirebaseAdmin/Auth/UserMetadata.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            this.CreationTimestamp = creationTimestamp;
+            this.LastSignInTimestamp = lastSignInTimestamp;
+        }
+
+        /// <summary>
+        /// Gets or sets a timestamp representing the date and time that the account was created.
+        /// </summary>
+        [JsonProperty("creationTimestamp")]
+        public long CreationTimestamp { get; set; }
```
hiranya911: Remove setters.

FirebaseAdmin/FirebaseAdmin/Auth/UserRecord.cs
```diff
@@ -48,6 +111,31 @@ private set
...
+
+        /// <summary>
+        /// Gets a value indicating whether the user account is disabled or not.
+        /// </summary>
+        public bool Disabled => this.disabled;
+
+        /// <summary>
+        /// Gets a list of provider data for this user.
+        /// </summary>
+        public List<ProviderUserInfo> Providers => this.providers;
```
hiranya911: Should be `IEnumerable<IUserInfo>`

FirebaseAdmin/FirebaseAdmin/Auth/UserRecord.cs
```diff
@@ -120,5 +245,19 @@ private static string SerializeClaims(IReadOnlyDictionary<string, object> claims
         {
             return NewtonsoftJsonSerializer.Instance.Serialize(claims);
         }
+
+        private static ReadOnlyDictionary<string, object> ParseCustomClaims(string customClaims)
```
hiranya911: Return `IReadOnlyDictionary` from this method.

FirebaseAdmin/FirebaseAdmin/Auth/UserRecord.cs
```diff
@@ -120,5 +245,19 @@ private static string SerializeClaims(IReadOnlyDictionary<string, object> claims
...
+        {
+            if (string.IsNullOrEmpty(customClaims))
+            {
+                return new ReadOnlyDictionary<string, object>(new Dictionary<string, object>());
```
hiranya911: You can use regular `Dictionary` here. It implements `IReadOnlyDictionary`.

FirebaseAdmin/FirebaseAdmin/Auth/UserRecord.cs
```diff
@@ -24,20 +25,82 @@ namespace FirebaseAdmin.Auth
...
+
+                for (int i = 0; i < count; i++)
+                {
+                    this.providers.Add(new ProviderUserInfo(user.Providers[i]));
+                }
+            }
+
+            this.tokensValidAfterTimestamp = user.ValidSince * 1000;
+            this.userMetaData = new UserMetadata(user.CreatedAt, user.LastLoginAt);
+            this.customClaims = UserRecord.ParseCustomClaims(user.CustomClaims);
```
hiranya911: This doesn't seem to be exposed at the moment.

FirebaseAdmin/FirebaseAdmin/Auth/UserMetadata.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        /// <summary>
+        /// Gets or sets a timestamp representing the last time that the user has logged in.
+        /// </summary>
+        [JsonProperty("lastSignInTimestamp")]
+        public long LastSignInTimestamp { get; set; }
```
hiranya911: I wonder if these should be represented as `DateTime` objects. But it's ok to leave it as is for now.

FirebaseAdmin/FirebaseAdmin/Auth/IUserInfo.cs
renatoc8: I can do that, but wouldn't we be departing from the Java implementation if we renamed it to Uid (I would imagine the "Get" naming convention wouldn't make sense for a property)?

FirebaseAdmin/FirebaseAdmin/Auth/UserRecord.cs
renatoc8: I'm not sure I understand your comment on this one.

FirebaseAdmin/FirebaseAdmin/Auth/UserMetadata.cs
renatoc8: Yeah, I think it would be helpful, but we can accomplish that with a readonly property that converts the timestamp value to a DateTime object. That way we don't lose the underlying timestamp data.

FirebaseAdmin/FirebaseAdmin/Auth/UserRecord.cs
hiranya911: We need a public property that exposes this attribute. Please disregard if you already have that, and I just missed it somehow.

FirebaseAdmin/FirebaseAdmin/Auth/IUserInfo.cs
hiranya911: Properties are more idiomatic in C#. So we should have something like this:

```
string Uid { get; }
string Email { get; }
// and so on...
```

FirebaseAdmin/FirebaseAdmin/Auth/UserRecord.cs
renatoc8: Yeah, those properties do exist:
`
/// <summary>
/// Gets a timestamp representing the date and time that this token will become active.
/// </summary>
public long TokensValidAfterTimestamp => this.tokensValidAfterTimestamp;

/// <summary>
/// Gets additional user metadata.
/// </summary>
public UserMetadata UserMetaData => this.userMetaData;

/// <summary>
/// Gets or sets the custom claims set on this user.
/// </summary>
[JsonIgnore]
public IReadOnlyDictionary<string, object> CustomClaims
{
	get => this.customClaims;
	set
	{
		CheckCustomClaims(value);
		this.customClaims = value;
	}
}
`
Assuming that's what you meant, I'll go ahead and mark this conversation as resolved then.

renatoc8:
@hiranya911 I believe I've marked all of the conversations as resolved, but I'm not sure how to mark the requested changes as complete. Is there anything else I have to do?

FirebaseAdmin/FirebaseAdmin/Auth/Internal/GetAccountInfoResponse.cs
```diff
@@ -0,0 +1,145 @@
...
+        /// <summary>
+        /// JSON data binding for user records.
+        /// </summary>
+        internal sealed class User
+        {
+            /// <summary>
+            /// Gets or sets the user's ID.
+            /// </summary>
+            [JsonProperty(PropertyName = "localId")]
+            public string UserID { get; set; }
```
hiranya911: UserId

FirebaseAdmin/FirebaseAdmin/Auth/Internal/GetAccountInfoResponse.cs
```diff
@@ -0,0 +1,145 @@
...
+        /// <summary>
+        /// JSON data binding for provider data.
+        /// </summary>
+        internal sealed class Provider
+        {
+            /// <summary>
+            /// Gets or sets the user's ID.
+            /// </summary>
+            [JsonProperty(PropertyName = "uid")]
+            public string UserID { get; set; }
```
hiranya911: UserId

hiranya911:
@renatoc8 can you also add a line to the `CHANGELOG` file at the root of this repo?

renatoc8:
Added the entry to CHANGELOG.

hiranya911:
This API is still under review. But I'll go ahead and merge the PR. If any changes are suggested by the review team, we can implement them separately.

PR #53, opened 2019-05-10
Adding the DeleteUser API and cleaning up the user management impl

### LINKED ISSUE #52: customAttributes size limitation is calculated with UTF-16, but it is UTF-8
If the byte-count is calculated with Unicode (UTF-16), it needs 2 bytes per character. The Java and Golang SDK use UTF-8 for that. This way these SDK's are able to store 1000 characters, whereas the dotnet SDK can only store 500 characters.

Can be found in **CheckCustomClaims**-method of **UserRecord**-class.

`
var customClaimsString = SerializeClaims(customClaims);
var byteCount = Encoding.Unicode.GetByteCount(customClaimsString);
if (byteCount > 1000)
{
    throw new ArgumentException($"Claims have to be not greater than 1000 bytes when serialized");
}
`

hiranya911:
1. Exposed `DeleteUserAsync()` method from the public API.
2. Added a skeleton implementation of the `UserArgs` class. Cleaned up the corresponding mutable operations from the `UserRecord` class.
3. Fixed a bug in how identity provider information is parsed (`uid` field renamed to `rawId`).
4. Updated and cleaned up unit tests. Added an integration test to cover the new APIs.
5. Encoding custom claims as UTF8 instead of Unicode (resolves #52).


DenSmoke:
Great PR. I wanted this function in .NET Admin SDK for a long time. Is it possible to accelerate merge?

hiranya911:
Hi @DenSmoke. This is currently stuck in review, and the earliest we can release this is next week.

PR #58, opened 2019-05-15
Adding the ListUsers API

### LINKED ISSUE #51: Request all users
This SDK shall also implement a functionality to retrieve all users. This is already implemented in the other 3 languages and is a main function for an admin SDK.

ChristopherLenz:
Added the "ListUsersAsync"-method to "FirebaseAuth"-class (resolves #51  )


hiranya911:
I believe this API should be based on `Page` and `PagedEnumerable` classes from Google GAX.

https://googleapis.github.io/google-cloud-dotnet/docs/guides/page-streaming.html

ChristopherLenz:
> I believe this API should be based on `Page` and `PagedEnumerable` classes from Google GAX.
> 
> https://googleapis.github.io/google-cloud-dotnet/docs/guides/page-streaming.html

I changed the paging-implementation. The drawback is that the previous internal classes used for the JSON-response have to be public. ATM I let them in the internal-folder. If this (internal --> public) is okay ,I would clean this up.

hiranya911:
Thanks @ChristopherLenz. The public API for this functionality should be something like this:

```
public PagedEnumerable<ExportedUserRecords, ExportedUserRecord> ListUsers(string pageToken)
public PagedEnumerable<ExportedUserRecords, ExportedUserRecord> ListUsers(string pageToken, CancellationToken cancellationToken)
public PagedEnumerable<ExportedUserRecords, ExportedUserRecord> ListUsers(String pageToken, int maxResults)
public PagedEnumerable<ExportedUserRecords, ExportedUserRecord> ListUsers(String pageToken, int maxResults, CancellationToken cancellationToken)
```

`ExportedUserRecord` and `ExportedUserRecords` should be new public types, and they should be the only types that get added to the public API surface.

It might also be easier to implement this by extending from the [`RestPagedEnumerable`](https://github.com/googleapis/gax-dotnet/blob/master/Google.Api.Gax.Rest/RestPagedEnumerable.cs) instead of writing our own from the scratch.

Then there's the question of asynchrony. All public APIs in the SDK are currently asynchronous. Ideally we should keep it that way. In that case it makes more sense to implement the APIs to return `PagedAsyncEnumerable`. But I'm not sure.

ChristopherLenz:
> Thanks @ChristopherLenz. The public API for this functionality should be something like this:
> 
> ```
> public PagedEnumerable<ExportedUserRecords, ExportedUserRecord> ListUsers(string pageToken)
> public PagedEnumerable<ExportedUserRecords, ExportedUserRecord> ListUsers(string pageToken, CancellationToken cancellationToken)
> public PagedEnumerable<ExportedUserRecords, ExportedUserRecord> ListUsers(String pageToken, int maxResults)
> public PagedEnumerable<ExportedUserRecords, ExportedUserRecord> ListUsers(String pageToken, int maxResults, CancellationToken cancellationToken)
> ```

I added the pageToken, but removed the **maxResults** and **cancellationToken**. These parameters are used in subsequent method-calls on the pagedEnumerable itself.
```
var usersPage = firebaseAuth.ListUsersAsync(string.Empty);
var userPage = await usersPage.ReadPageAsync(10, source.Token)
```

> `ExportedUserRecord` and `ExportedUserRecords` should be new public types, and they should be the only types that get added to the public API surface.

I will do this if you are satisfied with all other tasks.

> It might also be easier to implement this by extending from the [`RestPagedEnumerable`](https://github.com/googleapis/gax-dotnet/blob/master/Google.Api.Gax.Rest/RestPagedEnumerable.cs) instead of writing our own from the scratch.

I did, but I'm not sure if the pageToken-handling is as expected. The **requestProvider** of **RestPagedAsyncEnumerable** is called for every request. This means that the pageToken is reset for every subsequent call on **ReadPageAsync**.
 
> Then there's the question of asynchrony. All public APIs in the SDK are currently asynchronous. Ideally we should keep it that way. In that case it makes more sense to implement the APIs to return `PagedAsyncEnumerable`. But I'm not sure.

Yes I made everything ~~a~~sync, because of the PagedEnumerable. I was not aware, that there is an async-version. Now it is async again.

hiranya911:
Thanks @ChristopherLenz. At a glance this looks correct. We just need the `ExportedUserRecord` types, some tests and clean up. I will take another thorough look, once those changes are in.

hiranya911:
Ok, having taken another look at other gcloud SDKs, I think our public API should be something like this:

```
public PagedAsyncEnumerable<ExportedUserRecords, ExportedUserRecord> ListUsersAsync(ListUsersOptions options)

class ListUsersOptions {
  public int? PageSize;
  public string PageToken;
}
```

For comparison with a similar API, see: https://googleapis.github.io/google-cloud-dotnet/docs/Google.Cloud.Storage.V1/api/Google.Cloud.Storage.V1.StorageClient.html#Google_Cloud_Storage_V1_StorageClient_ListBucketsAsync_System_String_Google_Cloud_Storage_V1_ListBucketsOptions_

ChristopherLenz:
Thanks for the tips @hiranya911. I've added all the requests you made. I don't know if I will find time to add some tests today.

FirebaseAdmin/FirebaseAdmin/Auth/ExportedUserRecord.cs
```diff
@@ -0,0 +1,95 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace FirebaseAdmin.Auth
+{
+    /// <summary>
+    /// Contains metadata associated with a Firebase user account.
+    /// </summary>
+    public sealed class ExportedUserRecord
```
hiranya911: Just extend the `UserRecord` class (you will have to unseal it first). Also add the `PasswordHash` and `PasswordSalt` fields.

FirebaseAdmin/FirebaseAdmin/Auth/ExportedUserRecord.cs
```diff
@@ -0,0 +1,95 @@
...
+                PhotoUrl = userRecord.PhotoUrl,
+                Disabled = userRecord.Disabled,
+                CreatedAt = userRecord.CreatedAt,
+                LastLoginAt = userRecord.LastLoginAt,
+                ValidSince = userRecord.ValidSince,
+                CustomClaims = userRecord.CustomClaims,
+            };
+        }
+    }
+}
```
hiranya911: Add newline to end of file. Here and elsewhere.

FirebaseAdmin/FirebaseAdmin/Auth/ExportedUserRecord.cs
```diff
@@ -0,0 +1,95 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace FirebaseAdmin.Auth
+{
+    /// <summary>
+    /// Contains metadata associated with a Firebase user account.
```
hiranya911: Needs a bit more details. Look at the descriptions used in the Java SDK, and reuse the text as needed.

FirebaseAdmin/FirebaseAdmin/Auth/ExportedUserRecords.cs
```diff
@@ -0,0 +1,34 @@
...
+// limitations under the License.
+
+using System.Collections.Generic;
+
+namespace FirebaseAdmin.Auth
+{
+    /// <summary>
+    /// Contains a collection of Firebase user accounts.
+    /// </summary>
+    public sealed class ExportedUserRecords
```
hiranya911: We probably want a better API here. But this is a good start.

FirebaseAdmin/FirebaseAdmin/Auth/UserRecordPageManager.cs
```diff
@@ -0,0 +1,47 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System.Collections.Generic;
+using Google.Api.Gax.Rest;
+
+namespace FirebaseAdmin.Auth
+{
+    internal class UserRecordPageManager : IPageManager<UserRecordServiceRequest, ExportedUserRecords, ExportedUserRecord>
```
hiranya911: Lets call this `ListUsersPageManager`.

FirebaseAdmin/FirebaseAdmin/Auth/UserRecordServiceRequest.cs
```diff
@@ -0,0 +1,181 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Google.Apis.Discovery;
+using Google.Apis.Json;
+using Google.Apis.Requests;
+using Google.Apis.Services;
+
+namespace FirebaseAdmin.Auth
+{
+    internal class UserRecordServiceRequest : IClientServiceRequest<ExportedUserRecords>
```
hiranya911: `ListUsersRequest`

FirebaseAdmin/FirebaseAdmin/Auth/ExportedUserRecord.cs
```diff
@@ -0,0 +1,97 @@
...
+            this.lastLoginAt = user.LastLoginAt;
+            this.validSince = user.ValidSince;
+            this.passwordHash = user.PasswordHash;
+            this.passwordSalt = user.PasswordSalt;
+        }
+
+        /// <summary>
+        /// Gets the timestamp representing the time that the user account was created.
+        /// </summary>
+        public long CreatedAt
```
hiranya911: This class should only expose PasswordHash and PasswordSalt. Everything else should be inherited from UserRecord.

FirebaseAdmin/FirebaseAdmin/Auth/FirebaseUserManager.cs
```diff
@@ -98,6 +99,11 @@ public static FirebaseUserManager Create(FirebaseApp app)
             return new UserRecord(user);
         }
 
+        public ListUsersRequest CreateUserRecordServiceRequest(ListUsersOptions requestOptions)
```
hiranya911: CreateListUserRequest

ChristopherLenz:
@hiranya911 is there any work to do for me?

hiranya911:
Hi @ChristopherLenz. This is still lacking tests for some of the functionality. Ideally we need to test that iterating over users via `PagedAsyncEnumerator` works as expected.

ChristopherLenz:
> Hi @ChristopherLenz. This is still lacking tests for some of the functionality. Ideally we need to test that iterating over users via `PagedAsyncEnumerator` works as expected.

Okay I added a test for that. Please let me know if it is okay now.

FirebaseAdmin/FirebaseAdmin.Tests/Auth/FirebaseUserManagerTest.cs
```diff
@@ -118,6 +121,147 @@ public async Task GetUserByIdUserNotFound()
                 async () => await userManager.GetUserById("user1"));
         }
 
+        [Fact]
+        public async Task ListUsersPaged()
+        {
+            var nextPageToken = Guid.NewGuid().ToString();
+            var handlerInit = new MockMessageHandler()
```
hiranya911: Lets call this `firstCallHandler` and the next one `secondCallHandler`.

FirebaseAdmin/FirebaseAdmin.Tests/Auth/FirebaseUserManagerTest.cs
```diff
@@ -118,6 +121,147 @@ public async Task GetUserByIdUserNotFound()
...
+            var userManager = new FirebaseUserManager(
+                new FirebaseUserManagerArgs
+                {
+                    Credential = MockCredential,
+                    ProjectId = MockProjectId,
+                    ClientFactory = factory,
+                });
+
+            var requestOptions = new ListUsersOptions();
+            var usersPage = new RestPagedAsyncEnumerable<ListUsersRequest, ExportedUserRecords, ExportedUserRecord>(
```
hiranya911: Can we push all this into a method in `FirebaseUserManager`?

```
PagedAsyncEnumerable<...> ListUsers(ListUsersOptions options)
{
  ....
}
```

FirebaseAdmin/FirebaseAdmin/Auth/UserRecord.cs
```diff
@@ -177,6 +184,30 @@ public string ProviderId
         /// </summary>
         public long TokensValidAfterTimestamp => this.tokensValidAfterTimestamp;
 
+        /// <summary>
+        /// Gets the timestamp representing the time that the user account was created.
+        /// </summary>
+        public long CreatedAt
```
hiranya911: This and the next property should be exposed as `UserMetadata`. The master branch already contains the change. Just sync up your branch.

ChristopherLenz:
@hiranya911 
So I think that's all now :-)

FirebaseAdmin/FirebaseAdmin/Auth/UserRecord.cs
```diff
@@ -120,6 +123,21 @@ internal UserRecord(GetAccountInfoResponse.User user)
         /// </summary>
         public IUserInfo[] ProviderData { get; }
 
+        /// <summary>
```
hiranya911: Sorry if I wasn't clear in my earlier review. But these properties are not needed. CreatedAt and LastLoginAt are already in UserMetadata and ValidSince is same as TokensValidAfterTimestamp

PR #67, opened 2019-06-03
Implementing the CreateUserAsync() API

hiranya911:
* Renamed `UserArgs` to `UserRecordArgs` and made it public.
* Added the create user API and related tests.
* Added `.ConfigureAwait(false)` to all async API calls.

go/dotnet-user-mgt


hiranya911:
@egilmorez please review the documentation updated in the `FirebaseAuth` class.

FirebaseAdmin/FirebaseAdmin/Auth/FirebaseAuth.cs
```diff
@@ -263,6 +263,41 @@ public async Task<FirebaseToken> VerifyIdTokenAsync(string idToken)
                 .ConfigureAwait(false);
         }
 
+        /// <summary>
+        /// Creates a new user account with the attributes contained in the specified <see cref="UserRecordArgs"/>.
+        /// </summary>
+        /// <param name="args">Attributes that will be added to the new user account.</param>
```
egilmorez: Just so you don't think I'm being arbitrary about "will" -- I think it helpfully disambiguates here. If we said "attributes of" or "attributes for," it might suggest that the attributes already exist for the account.  So "will be" works best here :)

PR #90, opened 2019-07-18
Introducing FirebaseMessagingException and FCM error handling support

hiranya911:
* Added `FirebaseMessagingClass` that extends from `FirebaseException`.
* Added `MessagingErrorCode` enum type.
* Added `HttpErrorHandler`, `PlatformErrorHandler` and `MessagingErrorHandler` classes for handling different types of error responses.


FirebaseAdmin/FirebaseAdmin/HttpErrorHandler.cs
```diff
@@ -0,0 +1,107 @@
...
+                Message = message,
+            };
+        }
+
+        protected virtual FirebaseException CreateException(FirebaseExceptionArgs args)
+        {
+            return new FirebaseException(args.Code, args.Message, response: args.HttpResponse);
+        }
+
+        internal sealed class ErrorInfo
```
bklimt: `FirebaseException`, `ErrorInfo`, and `FirebaseExceptionArgs` all have very similar data. Could we combine a couple of these? 

Could we get rid of `ErrorInfo` and have `ExtractErrorInfo` just return a `FirebaseExceptionArgs` instead? I would say the other fields could be unset, but it looks like it has the data for the other two fields already.

FirebaseAdmin/FirebaseAdmin/Messaging/FirebaseMessagingClient.cs
```diff
@@ -239,6 +234,20 @@ private async Task<HttpResponseMessage> SendRequestAsync(object body, Cancellati
             return batch;
         }
 
+        private async Task<FirebaseMessagingException> CreateExceptionFor(HttpResponseMessage response)
```
bklimt: That 'For' at the end of the name is really weird in C#.

FirebaseAdmin/FirebaseAdmin/HttpErrorHandler.cs
```diff
@@ -0,0 +1,107 @@
...
+            {
+                Code = info.Code,
+                Message = info.Message,
+                HttpResponse = response,
+                ResponseBody = body,
+            };
+            throw this.CreateException(args);
+        }
+
+        protected virtual ErrorInfo ExtractErrorInfo(HttpResponseMessage response, string body)
```
bklimt: These two methods could really use a comment, since it seems that part of the contract for this class is that other classes should subclass it and override these two methods.

FirebaseAdmin/FirebaseAdmin/Messaging/FirebaseMessagingClient.cs
```diff
@@ -157,7 +152,7 @@ internal static string ClientVersion
             }
             catch (HttpRequestException e)
             {
-                throw new FirebaseException("Error while calling the FCM service.", e);
+                throw e.ToFirebaseException();
```
bklimt: It's a little weird to me that the messaging client won't always throw an instance of `FirebaseMessagingException`. I mean, I understand that there's no extra information to be included. But it seems like people might expect it to always throw the same type. And `FirebaseMessagingException` does support the messaging code being null. Thoughts?

FirebaseAdmin/FirebaseAdmin/Messaging/SendResponse.cs
```diff
@@ -40,7 +40,7 @@ private SendResponse(FirebaseException exception)
         /// <summary>
         /// Gets an exception if the send operation failed. Otherwise returns null.
         /// </summary>
-        public FirebaseException Exception { get; }
+        public FirebaseMessagingException Exception { get; }
```
bklimt: I know this isn't wrong, but it also reinforces what I was saying earlier. `SendAsync` can throw non-`FirebaseMessagingException` `FirebaseException`s, but when a send is part of `SendAllAsync`, a send can only fail with a `FirebaseMessagingException`. I think the set of `Exception` types that `SendAsync` and an individual send in a `SendAllAsync` can throw should probably be the same.

FirebaseAdmin/FirebaseAdmin/PlatformErrorHandler.cs
```diff
@@ -0,0 +1,89 @@
...
+// limitations under the License.
+
+using System.Collections.Generic;
+using System.Net.Http;
+using Google.Apis.Json;
+using Newtonsoft.Json;
+
+namespace FirebaseAdmin
+{
+    internal class PlatformErrorHandler : HttpErrorHandler
```
bklimt: This class could use a comment explaining how it fits into the overall picture.

FirebaseAdmin/FirebaseAdmin/HttpErrorHandler.cs
hiranya911: Done. Removed `ErrorInfo` altogether and using `FirebaseExceptionArgs` in its place.

FirebaseAdmin/FirebaseAdmin/Messaging/FirebaseMessagingClient.cs
hiranya911: Removed.

FirebaseAdmin/FirebaseAdmin/HttpErrorHandler.cs
hiranya911: Done

FirebaseAdmin/FirebaseAdmin/Messaging/FirebaseMessagingClient.cs
hiranya911: Good point. Actually I meant to throw only `FirebaseMessagingException` from this API, but missed that requirement in the implementation. I've fixed it now. Also updated the API reference in `FirebaseMessaging` and added a test case.

FirebaseAdmin/FirebaseAdmin/Messaging/SendResponse.cs
hiranya911: Done. FCM APIs now only throw `FirebaseMessagingException`.

FirebaseAdmin/FirebaseAdmin/PlatformErrorHandler.cs
hiranya911: Done

hiranya911:
Thanks @bklimt. I've made the suggested changes. Ready for another look.

bklimt:
LGTM

PR #99, opened 2019-08-05
Sending X-Client-Version header with auth requests

hiranya911:


bojeil-google:
Logic looks good to me.

PR #109, opened 2019-08-30
fix: Add .ConfigureAwait(false) to all async call

### LINKED ISSUE #108: Got Deadlock on FirebaseMessaging.SendAsync(msg).ConfigureAwait(false).GetAwaiter().GetResult()
I've got deadlock on FirebaseMessaging.SendAsync(msg).ConfigureAwait(false).GetAwaiter().GetResult()

i've investigate that lib doesn't do .ConfigureAwait(false) for all async call.

i've download source and solve it. it work fine

NiCoTinEz:
RELEASE NOTE: Fixed an issue that caused async API calls to deadlock in some environments.

Resolves #108


NiCoTinEz:
@googlebot I fixed it.

PR #119, opened 2019-09-27
Remove (base64) 'REDACTED' passwords from user records.

rsgowman:
These values *look* like password hashes, but aren't, leading to
potential confusion.

Additionally, added docs to CONTRIBUTING.md detailing how to add the
permission that causes password hashes to be properly returned as well
as adjusting the test failure message should the developer not add that
permission.

b/141189502

### DIFF
--- a/FirebaseAdmin/FirebaseAdmin.IntegrationTests/FirebaseAuthTest.cs
+++ b/FirebaseAdmin/FirebaseAdmin.IntegrationTests/FirebaseAuthTest.cs
@@ -324,8 +324,15 @@ namespace FirebaseAdmin.IntegrationTests
                     if (users.Contains(uid) && !listedUsers.Contains(uid))
                     {
                         listedUsers.Add(uid);
-                        Assert.NotNull(enumerator.Current.PasswordHash);
-                        Assert.NotNull(enumerator.Current.PasswordSalt);
+                        var errMsgTemplate = "Missing {0} field. A common cause would be "
+                            + "forgetting to add the 'Firebase Authentication Admin' permission. "
+                            + "See instructions in CONTRIBUTING.md";
+                        AssertWithMessage.NotNull(
+                            enumerator.Current.PasswordHash,
+                            errMsgTemplate.Format("PasswordHash"));
+                        AssertWithMessage.NotNull(
+                            enumerator.Current.PasswordSalt,
+                            errMsgTemplate.Format("PasswordSalt"));
                     }
                 }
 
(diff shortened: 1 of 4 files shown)


FirebaseAdmin/FirebaseAdmin.IntegrationTests/FirebaseAuthTest.cs
```diff
@@ -324,8 +324,15 @@ public async Task ListUsers()
                     if (users.Contains(uid) && !listedUsers.Contains(uid))
                     {
                         listedUsers.Add(uid);
-                        Assert.NotNull(enumerator.Current.PasswordHash);
-                        Assert.NotNull(enumerator.Current.PasswordSalt);
+                        var errMsgTemplate = "Missing {0} field. A common cause would be "
```
rsgowman: It feels more natural to me to just inline the error message (despite some slight duplication). However, that runs afoul of stylecop, so a `#pragma warning disable` would be required. So I just went with the errMsgTemplate variable as seen here.

Using the `#pragma`, this block would look like this:
```c#
#pragma warning disable SA1118 // ParameterMustNotSpanMultipleLines
AssertWithMessage.NotNull(
    enumerator.Current.PasswordHash,
    "Missing PasswordHash field. A common cause would be forgetting to " 
    + "add the 'Firebase Authentication Admin' permission. See " 
    + "instructions in CONTRIBUTING.md");
AssertWithMessage.NotNull(
    enumerator.Current.PasswordSalt,
    "Missing PasswordSalt field. A common cause would be forgetting to " 
    + "add the 'Firebase Authentication Admin' permission. See " 
    + "instructions in CONTRIBUTING.md");
#pragma warning restore SA1118 // ParameterMustNotSpanMultipleLines
```

FirebaseAdmin/FirebaseAdmin.Tests/Auth/ExportedUserRecordTest.cs
```diff
@@ -151,5 +151,18 @@ public void AllProperties()
...
             Assert.Equal(UserRecord.UnixEpoch.AddMilliseconds(150), metadata.LastSignInTimestamp);
         }
+
+        [Fact]
+        public void RedactedPasswordCleared()
+        {
+            var user = new ExportedUserRecord(new GetAccountInfoResponse.User()
+            {
+                UserId = "user1",
+                PasswordHash = System.Convert.ToBase64String(
```
hiranya911:
```
using System;
using System.Text;
```

FirebaseAdmin/FirebaseAdmin.IntegrationTests/FirebaseAuthTest.cs
```diff
@@ -408,4 +415,18 @@ internal static RandomUser Create()
             };
         }
     }
+
+    /**
+     * Additional Xunit style asserts that allow specifying an error message upon failure.
+     */
+    internal class AssertWithMessage
```
hiranya911: May be make this a private static class to further reduce the visibility? I don't feel too strongly either way. So your call.

PR #127, opened 2019-10-20
Implement feature for importing users

levimatheri:
I have implemented just the base API here that sets the properties as well as the `UserImportHash`


billybraga:
@levimatheri Was/Is this implemented elsewhere?

levimatheri:
@billybraga Yes. Java SDK has this implemented. I just haven't had the time to port it over. Want to take a stab at it?

billybraga:
@levimatheri Yes, maybe.

PR #141, opened 2020-02-02
fix(auth): Added EMAIL_EXISTS Auth error message.

kcieslinski:
Added missing EMAIL_EXISTS Auth error message.
Not sure if the DUPLICATE_EMAIL message is already deprecated, as firebase api documentation is not clear about this for me.

RELEASE NOTE: User management APIs now correctly raise `ErrorCode.AlreadyExists` when an already in-use email address is specified for a user.


hiranya911:
Thanks @kcieslinski. We need to handle both error codes. See https://github.com/firebase/firebase-admin-python/pull/348 for some more context.

PR #172, opened 2020-04-15
feat(auth): Added revocation check support for VerifyIdTokenAsync()

hiranya911:
* Added `RevokedIdToken` auth error code.
 * Added `VerifyIdTokenAsync(string, bool)` overload.
* Moved ID token verification tests to a new `IdTokenVerificationTest` class.
* Changed the structure of these tests, so that they are invoked via `FirebaseAuth` API instead of the internal API.


FirebaseAdmin/FirebaseAdmin.Tests/Auth/IdTokenVerificationTest.cs
```diff
@@ -0,0 +1,466 @@
+// Copyright 2018, Google Inc. All rights reserved.
```
lahirumaramba, line 1: Is this a new file? Should this be 2020?

hiranya911: Done

Issue #185, opened 2020-05-21
FR: Implement ImportUsersAsync() API


hiranya911:
Implement an API for importing Firebase Auth user accounts in bulk. Partial API proposal:

```
class FirebaseAuth
    Task<UserImportResult> ImportUsersAsync(IEnumerable<ImportUserRecordArgs> users)
    Task<UserImportResult> ImportUsersAsync(IEnumerable<ImportUserRecordArgs> users, UserImportOptions options)

class ImportUserRecordArgs
    string Uid { get; set; }
    string Email { get; set; }
    boolean EmailVerified { get; set; }
    string DisplayName { get; set; }
    string PhoneNumber { get; set; }
    string PhotoUrl { get; set; }
    boolean Disabled { get; set; }
    UserMetadata UserMetadata { get; set; }
    byte[] PasswordHash { get; set; }
    byte[] PasswordSalt { get; set; }
    IEnumerable<UserProvider> UserProviders { get; set; }
    IReadOnlyDictionary<string, object> CustomClaims  { get; set; }

class UserImportResult
    int SuccessCount { get; }
    int FailureCount { get; }
    IReadOnlyList<ErrorInfo> Errors { get; }
```

MathBunny:
Merged in #191, remaining work is to make a release that includes the change.

hiranya911:
Will be included in the next release.

PR #191, opened 2020-06-01
feat(auth): Add bulk user import functionality and hashing algorithms namespace

MathBunny:
## Summary
This pull request adds the ability to bulk-import users into Firebase, resolving #185. 

```
class FirebaseAuth
    Task<UserImportResult> ImportUsersAsync(IEnumerable<ImportUserRecordArgs> users)
    Task<UserImportResult> ImportUsersAsync(IEnumerable<ImportUserRecordArgs> users, UserImportOptions options)

class ImportUserRecordArgs
    string Uid { get; set; }
    string Email { get; set; }
    boolean EmailVerified { get; set; }
    string DisplayName { get; set; }
    string PhoneNumber { get; set; }
    string PhotoUrl { get; set; }
    boolean Disabled { get; set; }
    UserMetadata UserMetadata { get; set; }
    byte[] PasswordHash { get; set; }
    byte[] PasswordSalt { get; set; }
    IEnumerable<UserProvider> UserProviders { get; set; }
    IReadOnlyDictionary<string, object> CustomClaims  { get; set; }

class UserImportResult
    int SuccessCount { get; }
    int FailureCount { get; }
    IReadOnlyList<ErrorInfo> Errors { get; }
```

It also defines the `.auth.hash` namespace, used for specifying different hashing algorithms. 

The code aims to conform with the internally discussed API proposal and be consistent with the other language implementations while being idiomatic to C#.

## Test Plan
- Unit tests
   - All unit tests pass
- Integration tests
    - E2E tests verifying the validation and serialization of objects and REST call to the backend returns successfully
    - All integration tests pass
    - Performed visual inspection on the Firebase console to ensure that users show up along with the identity providers appearing (no other manual testing beyond that was performed)

---

## Remarks
Several things I am not sure about:
- Using `int?` vs `int = Int32.MinValue` for determining if a field was set previously
    - Uncertain if the `Nullable<T>` wrapper creates any performance overhead, couldn't find anything online supporting this claim
- If derived key length, block sizes, etc. can be zero
- If there's a way to verify that tests succeeded beyond the REST API indicating success
    - Ex: intercepting the serialized object before getting sent to the post request via mocking
- `this.EmailVerified` and several other fields default to `false` in the Java implementation if they were not set before, but here I chose to not include them as part of the payload


FirebaseAdmin/FirebaseAdmin/Auth/UserRecordArgs.cs
```diff
@@ -265,6 +242,29 @@ private static string CheckCustomClaims(IReadOnlyDictionary<string, object> clai
...
+
+        private static string CheckPassword(string password)
+        {
+            if (password != null)
+            {
+                if (password.Length < 6)
+                {
+                    throw new ArgumentException("Password must be at least 6 characters long.");
+                }
+            }
```
MathBunny, line 263: I wonder if instead the conditional can be:

```
if (password != null && password.Length < 6) 
{
    throw new ArgumentException("Password must be at least 6 characters long.");
}
```

FirebaseAdmin/FirebaseAdmin/Auth/Hash/Scrypt.cs
```diff
@@ -0,0 +1,148 @@
...
+
+        /// <summary>
+        /// Gets the hash name which is SCRYPT.
+        /// </summary>
+        protected override string HashName { get { return "SCRYPT"; } }
+
+        /// <summary>
+        /// Gets the minimum number of rounds for a Scrypt hash which is 0.
+        /// </summary>
+        protected override int MinRounds { get { return 1; } }
```
lahirumaramba, line 120: Should this be `return 0;`? Or should the comment be set to `1`?

FirebaseAdmin/FirebaseAdmin/Auth/UserRecordArgs.cs
lahirumaramba: I think you can. `&&` will short-circuit and will not evaluate the second operand if password is `null`.

lahirumaramba: Are `null` passwords allowed here?

MathBunny: I think that null passwords can be passed into the `UserRecordArgs` constructor, maybe for a user without a password

I didn't write this chunk of code, I just happened to come across it when I was shuffling it to address linter comments and thought I'd add the `&&` to the ifs :) it's not very related to my actual changes otherwise

FirebaseAdmin/FirebaseAdmin/Auth/Hash/Scrypt.cs
MathBunny: Good catch, it should be `return 0`. Thanks!

FirebaseAdmin/FirebaseAdmin/Auth/UploadAccountResponse.cs
```diff
@@ -0,0 +1,31 @@
...
+
+using System.Collections.Generic;
+using Newtonsoft.Json;
+
+namespace FirebaseAdmin.Auth
+{
+    /// <summary>
+    /// Represents the response from identity toolkit for a user import request.
+    /// </summary>
+    public class UploadAccountResponse
```
hiranya911, line 23: internal sealed

Actually, consider just nesting this in `FirebaseUserManager`. Doesn't look like we need a top-level class for such a small thing.

FirebaseAdmin/FirebaseAdmin/Auth/FirebaseAuth.cs
```diff
@@ -720,6 +720,96 @@ public async Task<DeleteUsersResult> DeleteUsersAsync(IReadOnlyList<string> uids
                 .ConfigureAwait(false);
         }
 
+        /// <summary>
+        /// Imports the provided list of users into Firebase Auth. You can import a maximum of
+        /// 1000 users at a time. This operation is optimized for bulk imports and does not
+        /// check identifier uniqueness which could result in duplications.</summary>
+        /// <p><a cref="UserImportOptions">UserImportOptions</a> is required to import users with
```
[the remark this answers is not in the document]
MathBunny: Sorry, I am a little unsure, should I add a newline before line 727? Or should I use the `<para>` tag and nest that line within the summary?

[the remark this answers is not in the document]
hiranya911:
```
/// <summary>
/// Text.
///
/// <para>Text</para>
/// </summary>
```

MathBunny:
Thanks for the feedback @hiranya911  @lahirumaramba, PTAL :)

FirebaseAdmin/FirebaseAdmin/Auth/ImportUserRecordArgs.cs
```diff
@@ -0,0 +1,210 @@
...
+        internal sealed class ImportUserRequest
+        {
+            internal ImportUserRequest(ImportUserRecordArgs args)
+            {
+                this.Uid = UserRecordArgs.CheckUid(args.Uid, true);
+
+                if (!string.IsNullOrEmpty(args.Email))
+                {
+                    UserRecordArgs.CheckEmail(args.Email);
+                    this.Email = args.Email;
```
hiranya911, line 119: These can be made one liners:

```
this.Email = UserRecordArgs.CheckEmail(args.Email);
```

FirebaseAdmin/FirebaseAdmin/Auth/UserImportHash.cs
```diff
@@ -0,0 +1,77 @@
...
+    /// </remarks>
+    public abstract class UserImportHash
+    {
+        private string hashName;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="UserImportHash"/> class.
+        /// </summary>
+        /// <param name="hashName">The name of the hashing algorithm.</param>
+        public UserImportHash(string hashName)
```
hiranya911, line 39: protected

FirebaseAdmin/FirebaseAdmin/Auth/UserImportHash.cs
```diff
@@ -0,0 +1,77 @@
...
+
+            properties.Add("hashAlgorithm", this.hashName);
+            return properties;
+        }
+
+        /// <summary>
+        /// Retrives dictionary with the specified properties of the hashing algorithm.
+        /// </summary>
+        /// <returns>Dictionary containing the specified properties of the hashing algorithm.</returns>
+        protected abstract IReadOnlyDictionary<string, object> GetOptions();
```
hiranya911, line 75: So we have `GetProperties()` and `GetOptions()` methods in this class, which are not very good names. How about we rename GetOptions to `GetHashConfiguration()` or something similar?

FirebaseAdmin/FirebaseAdmin/Auth/ImportUserRecordArgs.cs
```diff
@@ -0,0 +1,210 @@
...
+            return new ImportUserRequest(this);
+        }
+
+        private static string UrlSafeBase64Encode(byte[] bytes)
+        {
+            var base64Value = Convert.ToBase64String(bytes);
+            return base64Value.TrimEnd('=').Replace('+', '-').Replace('/', '_');
+        }
+
+        internal sealed class ImportUserRequest
```
hiranya911, line 110: Let's find a different name. You also have an `UserImportRequest` class.

FirebaseAdmin/FirebaseAdmin/Auth/UserImportOptions.cs
```diff
@@ -0,0 +1,46 @@
...
+using System;
+using System.Collections.Generic;
+
+namespace FirebaseAdmin.Auth
+{
+    /// <summary>
+    /// A collection of options that can be passed to the
+    /// <a cref="o:FirebaseAuth.ImportUsersAsync">FirebaseAuth.ImportUsersAsync</a> API.
+    /// </summary>
+    public class UserImportOptions
```
hiranya911, line 24: sealed

Should be used on pretty much all classes, unless there's a specific reason not to.

FirebaseAdmin/FirebaseAdmin/Auth/UserImportRequest.cs
```diff
@@ -0,0 +1,92 @@
...
+
+            if (usersToImport.Count() > MaxImportUsers)
+            {
+                throw new ArgumentException($"users list must not contain more than"
+                    + $" {MaxImportUsers} items");
+            }
+
+            bool hasPassword = false;
+            this.users = usersToImport.Select((user) => user.ToImportUserRequest());
+            hasPassword = usersToImport.Any((user) => user.HasPassword());
```
hiranya911, line 60: var hasPassword =

FirebaseAdmin/FirebaseAdmin/Auth/UserImportRequest.cs
```diff
@@ -0,0 +1,92 @@
...
+            UserImportOptions options)
+        {
+            if (usersToImport.Count() == 0)
+            {
+                throw new ArgumentException("users must not be empty");
+            }
+
+            if (usersToImport.Count() > MaxImportUsers)
+            {
+                throw new ArgumentException($"users list must not contain more than"
```
hiranya911, line 54: Nit: You don't need the `$` here now.

FirebaseAdmin/FirebaseAdmin/Auth/UserImportRequest.cs
```diff
@@ -0,0 +1,92 @@
...
+                    + $" {MaxImportUsers} items");
+            }
+
+            bool hasPassword = false;
+            this.users = usersToImport.Select((user) => user.ToImportUserRequest());
+            hasPassword = usersToImport.Any((user) => user.HasPassword());
+
+            if (hasPassword)
```
hiranya911, line 62:
```
if (hasPassword && options?.Hash == null)
```

FirebaseAdmin/FirebaseAdmin/Auth/UserImportRequest.cs
```diff
@@ -0,0 +1,92 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Gets or sets JsonExtensionData for putting hashing properties at the root of the
+        /// Json serialized object.
+        /// </summary>
+        /// <returns>Dictionary containing key/values for password hashing algorithm.</returns>
+        [JsonExtensionData]
+        protected Dictionary<string, object> HashProperties { get; set; }
```
hiranya911, line 81: Make the setter private.

FirebaseAdmin/FirebaseAdmin/Auth/UserImportRequest.cs
```diff
@@ -0,0 +1,92 @@
...
+    /// <summary>
+    /// Encapsulates user import requests by specifying hashing properties for passwords and
+    /// the list of users to be imported.
+    /// </summary>
+    internal class UserImportRequest
+    {
+        internal const int MaxImportUsers = 1000;
+
+        [JsonProperty("users")]
+        private readonly IEnumerable<ImportUserRecordArgs.ImportUserRequest> users;
```
hiranya911, line 31: I suppose this can also be a property with a private setter?

FirebaseAdmin/FirebaseAdmin/Auth/UserImportResult.cs
```diff
@@ -0,0 +1,62 @@
...
+    {
+        private readonly int users;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="UserImportResult"/> class based on supplied
+        /// users and <a cref="FirebaseUserManager.UploadAccountResponse">UploadAccountResponse</a> objects.
+        /// </summary>
+        /// <param name="users"> The number of users.</param>
+        /// <param name="response"> The UploadAccountResponse generated from the post request.</param>
+        internal UserImportResult(int users, FirebaseUserManager.UploadAccountResponse response)
```
hiranya911, line 33: Perhaps this constructor can accept `IEnumerable<ErrorInfo>` instead of `UploadAccountResponse`?

FirebaseAdmin/FirebaseAdmin/Auth/UserImportResult.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// </summary>
+        public IReadOnlyList<ErrorInfo> Errors { get; private set; }
+
+        /// <summary>
+        /// Gets the number of users that were imported successfully.
+        /// </summary>
+        /// <returns>Number of users successfully imported (possibly zero).</returns>
+        public int SuccessCount
+        {
+            get => this.users - this.Errors.Count;
```
hiranya911, line 50: `this.users - this.FailureCount`

FirebaseAdmin/FirebaseAdmin/Auth/Hash/Bcrypt.cs
```diff
@@ -0,0 +1,45 @@
...
+    public sealed class Bcrypt : UserImportHash
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Bcrypt"/> class.
+        /// Defines name to be equal to BCRYPT.
+        /// </summary>
+        public Bcrypt()
+            : base("BCRYPT") { }
+
+        // protected override string HashName { get { return "BCRYPT"; } }
```
hiranya911, line 32: Please remove.

FirebaseAdmin/FirebaseAdmin/Auth/Hash/Hmac.cs
```diff
@@ -0,0 +1,58 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace FirebaseAdmin.Auth.Hash
+{
+    /// <summary>
+    /// Base class for Hmac type hashes.
+    /// </summary>
+    public abstract class Hmac : UserImportHash
```
hiranya911, line 24: Can this be internal or protected?

FirebaseAdmin/FirebaseAdmin/Auth/Hash/RepeatableHash.cs
```diff
@@ -0,0 +1,90 @@
...
+using System;
+using System.Collections.Generic;
+
+namespace FirebaseAdmin.Auth.Hash
+{
+    /// <summary>
+    /// An abstract <a cref="UserImportHash">UserImportHash</a> implementation for specifying a
+    /// <c>Rounds</c> count in a given range.
+    /// </summary>
+    public abstract class RepeatableHash : UserImportHash
```
hiranya911, line 24: internal

FirebaseAdmin/FirebaseAdmin/Auth/Hash/RepeatableHash.cs
```diff
@@ -0,0 +1,90 @@
...
+    {
+        private int? rounds;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="RepeatableHash"/> class.
+        /// Propogates the name to UserImportHash.
+        /// </summary>
+        /// <param name="hashName">The name of the hashing algorithm.</param>
+        public RepeatableHash(string hashName)
```
hiranya911, line 33: Definitely make the constructor protected.

FirebaseAdmin/FirebaseAdmin/Auth/Hash/RepeatableHash.cs
```diff
@@ -0,0 +1,90 @@
...
+            : base(hashName) { }
+
+        /// <summary>
+        /// Gets or sets the number of rounds for the repeatable hash. Verifies that the
+        /// specified <c>Rounds</c> are within the required bounds.
+        /// </summary>
+        public int Rounds
```
hiranya911, line 40: Keep this as a regular getter/setter without validation. Perform all validation in `GetOptions()`. Here and in other similar classes.

FirebaseAdmin/FirebaseAdmin/Auth/Hash/PbkdfSha1.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// Initializes a new instance of the <see cref="PbkdfSha1"/> class.
+        /// Defines the name of the hash to be equal to PBKDF_SHA1.
+        /// </summary>
+        public PbkdfSha1()
+            : base("PBKDF_SHA1") { }
+
+        /// <summary>
+        /// Gets the minimum number of rounds for a Pbkdf Sha1 hash which is 0.
+        /// </summary>
+        protected override int MinRounds { get { return 0; } }
```
hiranya911, line 33: `{ get => 0; }`

FirebaseAdmin/FirebaseAdmin/Auth/Hash/Scrypt.cs
```diff
@@ -0,0 +1,150 @@
...
+        /// Initializes a new instance of the <see cref="Scrypt"/> class.
+        /// Defines the name of the hash to be equal to SCRYPT.
+        /// </summary>
+        public Scrypt()
+            : base("SCRYPT") { }
+
+        /// <summary>
+        /// Gets or sets the signer key for the hashing algorithm.
+        /// </summary>
+        public string Key
```
hiranya911, line 45: Use a regular property and validate in `GetOptions`.

```
public string Key { get; set; }
```

hiranya911, line 45: Actually this should be `byte[]` isn't it?

FirebaseAdmin/FirebaseAdmin/Auth/Hash/Scrypt.cs
```diff
@@ -0,0 +1,150 @@
...
+
+                var plainTextBytes = System.Text.Encoding.UTF8.GetBytes(value);
+                this.key = UrlSafeBase64Encode(plainTextBytes);
+            }
+        }
+
+        /// <summary>
+        /// Gets or sets the salt separator for the hashing algorithm.
+        /// </summary>
+        public string SaltSeparator
```
hiranya911, line 72: `byte[]`

FirebaseAdmin/FirebaseAdmin/Auth/Hash/Scrypt.cs
```diff
@@ -0,0 +1,150 @@
...
+        protected override IReadOnlyDictionary<string, object> GetOptions()
+        {
+            var dict = new Dictionary<string, object>((Dictionary<string, object>)base.GetOptions());
+            dict.Add("signerKey", this.Key);
+            dict.Add("memoryCost", this.MemoryCost);
+            dict.Add("saltSeparator", this.SaltSeparator);
+            return dict;
+        }
+
+        private static string UrlSafeBase64Encode(byte[] bytes)
```
hiranya911, line 144: Was this duplicated elsewhere? If so push it to a common place like `JwtUtils`.

FirebaseAdmin/FirebaseAdmin/Auth/Hash/StandardScrypt.cs
```diff
@@ -0,0 +1,165 @@
...
+        /// Defines the name of the hash to be equal to STANDARD_SCRYPT.
+        /// </summary>
+        public StandardScrypt()
+            : base("STANDARD_SCRYPT") { }
+
+        /// <summary>
+        /// Gets or sets the derived key length for the hashing algorithm.
+        /// <remarks>The length cannot be negative.</remarks>
+        /// </summary>
+        public int DerivedKeyLength
```
hiranya911, line 45: Lets clean this all up by using regular properties and validating in `GetOptions`

FirebaseAdmin/FirebaseAdmin/Auth/UserImportRequest.cs
MathBunny: I'm wondering wouldn't this be different behavior? If `hasPassword` is true but the hash isn't null, then the hash properties are added in the original logic. I can't always add it because the `GetHashProperties` will throw if it is null.

So unless I would add another conditional structure below it to accommodate the hash being non-null, this would behave differently I think

FirebaseAdmin/FirebaseAdmin/Auth/UserImportRequest.cs
MathBunny: I think I can make the entire Dictionary private, it's only used for serialization purposes

FirebaseAdmin/FirebaseAdmin/Auth/ImportUserRecordArgs.cs
MathBunny: Would `ImportUserRecordArgsRequest` be too long?

FirebaseAdmin/FirebaseAdmin/Auth/UserImportRequest.cs
MathBunny: That makes sense, although I think the readonly modifier is lost this way

```
internal IEnumerable<ImportUserRecordArgs.ImportUserRequest> Users { get; private set; }
```

FirebaseAdmin/FirebaseAdmin/Auth/Hash/Hmac.cs
MathBunny: Based on my attempt yesterday it looks like no, but the constructor can remain internal. This is due to "Inconsistent accessibility"

FirebaseAdmin/FirebaseAdmin/Auth/Hash/Bcrypt.cs
```diff
@@ -26,18 +26,16 @@ public sealed class Bcrypt : UserImportHash
         /// Initializes a new instance of the <see cref="Bcrypt"/> class.
         /// Defines name to be equal to BCRYPT.
         /// </summary>
-        public Bcrypt()
+        internal Bcrypt()
```
hiranya911, line 29: This needs to be public so that developers can instantiate it.

FirebaseAdmin/FirebaseAdmin/Auth/Hash/Hmac.cs
```diff
@@ -28,30 +27,30 @@ public abstract class Hmac : UserImportHash
...
-            if (string.IsNullOrEmpty(this.Key))
+            if (this.Key == null || this.Key.Length == 0)
             {
                 throw new ArgumentException("key must not be null or empty");
             }
 
             return new Dictionary<string, object>
             {
-                { "signerKey", Encoding.ASCII.GetBytes(this.Key) },
+                { "signerKey", this.Key },
```
hiranya911, line 53: Shouldn't you have to base64 encode the byte array? I think otherwise json marshalling will break down.

FirebaseAdmin/FirebaseAdmin/Auth/Hash/Md5.cs
```diff
@@ -24,17 +24,17 @@ public sealed class Md5 : RepeatableHash
         /// Initializes a new instance of the <see cref="Md5"/> class.
         /// Defines the name of the hash to be equal to MD5.
         /// </summary>
-        public Md5()
+        internal Md5()
```
hiranya911, line 27: public

Anything that we expect the developers to instantiate must be public. Only make the abstract base classes like `RepeatableHash` internal (or rather make their constructors internal).

FirebaseAdmin/FirebaseAdmin/Auth/Hash/PbkdfSha1.cs
```diff
@@ -24,17 +24,17 @@ public sealed class PbkdfSha1 : RepeatableHash
         /// Initializes a new instance of the <see cref="PbkdfSha1"/> class.
         /// Defines the name of the hash to be equal to PBKDF_SHA1.
         /// </summary>
-        public PbkdfSha1()
+        internal PbkdfSha1()
```
hiranya911, line 27: Same here.

FirebaseAdmin/FirebaseAdmin/Auth/Hash/RepeatableHash.cs
```diff
@@ -23,43 +23,18 @@ namespace FirebaseAdmin.Auth.Hash
...
     {
-        private int? rounds;
-
         /// <summary>
         /// Initializes a new instance of the <see cref="RepeatableHash"/> class.
         /// Propogates the name to UserImportHash.
         /// </summary>
         /// <param name="hashName">The name of the hashing algorithm.</param>
-        public RepeatableHash(string hashName)
+        internal RepeatableHash(string hashName)
```
hiranya911, line 31: Right 👍

FirebaseAdmin/FirebaseAdmin/Auth/Hash/Scrypt.cs
```diff
@@ -26,95 +26,27 @@ namespace FirebaseAdmin.Auth.Hash
...
-        private string saltSeparator;
-
-        private int? memoryCost;
-
         /// <summary>
         /// Initializes a new instance of the <see cref="Scrypt"/> class.
         /// Defines the name of the hash to be equal to SCRYPT.
         /// </summary>
-        public Scrypt()
+        internal Scrypt()
```
hiranya911, line 33: public

FirebaseAdmin/FirebaseAdmin/Auth/Hash/Scrypt.cs
```diff
@@ -130,21 +62,42 @@ public int MemoryCost
...
-        protected override IReadOnlyDictionary<string, object> GetOptions()
+        protected override IReadOnlyDictionary<string, object> GetHashConfiguration()
         {
-            var dict = new Dictionary<string, object>((Dictionary<string, object>)base.GetOptions());
+            if (this.Key == null)
+            {
+                throw new ArgumentNullException("key must be initialized");
+            }
+
+            if (this.Key.Length == 0)
```
hiranya911, line 74:
```
if (this.Key == null || this.Key.Length == 0)
```

FirebaseAdmin/FirebaseAdmin/Auth/Hash/Scrypt.cs
```diff
@@ -130,21 +62,42 @@ public int MemoryCost
...
+            {
+                throw new ArgumentException("key must not be null or empty");
+            }
+
+            if (this.SaltSeparator == null)
```
hiranya911, line 79: Make sure to base64 encode the byte arrays.

FirebaseAdmin/FirebaseAdmin/Auth/Hash/Scrypt.cs
```diff
@@ -130,21 +62,42 @@ public int MemoryCost
...
+            {
+                throw new ArgumentNullException("memory cost must be set");
+            }
+
+            if (this.MemoryCost < 1 || this.MemoryCost > 14)
+            {
+                throw new ArgumentException("memory cost must be between 1 and 14 (inclusive)");
+            }
+
+            var dict = new Dictionary<string, object>(
```
hiranya911, line 94: Avoid the cast:

```
var dict = new Dictionary<string, object>()
{
  // child class properties
};

for (var entry in base.GetHashConfiguration())
{
  dict[entry.Key] = entry.Value;
}
```

FirebaseAdmin/FirebaseAdmin/Auth/Hash/StandardScrypt.cs
```diff
@@ -23,143 +23,69 @@ namespace FirebaseAdmin.Auth.Hash
...
         /// Dictionary defining options such as derived key length, block size, parallization and memory cost.
         /// </returns>
-        protected override IReadOnlyDictionary<string, object> GetOptions()
+        protected override IReadOnlyDictionary<string, object> GetHashConfiguration()
         {
+            this.CheckPropertyNegativeOrZero(this.DerivedKeyLength, "DerivedKeyLength");
+            this.CheckPropertyNegativeOrZero(this.BlockSize, "BlockSize");
+            this.CheckPropertyNegativeOrZero(this.Parallelization, "Parallelization");
+            this.CheckPropertyNegativeOrZero(this.MemoryCost, "MemoryCost");
+
```
hiranya911, line 69: Use the inlined syntax for dictionary initialization (everywhere in this PR):

```
var dict = new Dictionary<string, object>()
{
  {"dkLen", this.DerivedKeyLength},
  // other properties...
};
```

FirebaseAdmin/FirebaseAdmin/Auth/Hash/StandardScrypt.cs
```diff
@@ -23,143 +23,69 @@ namespace FirebaseAdmin.Auth.Hash
...
             var dict = new Dictionary<string, object>();
             dict.Add("dkLen", this.DerivedKeyLength);
             dict.Add("blockSize", this.BlockSize);
             dict.Add("parallization", this.Parallelization);
             dict.Add("memoryCost", this.MemoryCost);
             return dict;
         }
+
+        private void CheckPropertyNegativeOrZero(int? value, string name)
```
hiranya911, line 78:
```
private static void ThrowIfNullOrNegative(int? value, string message)
{
    if (value == null || value < 0)
    {
        throw new ArgumentException(message);
    }
}
```

FirebaseAdmin/FirebaseAdmin/Auth/ImportUserRecordArgs.cs
```diff
@@ -96,40 +96,19 @@ internal bool HasPassword()
...
 
-        internal ImportUserRequest ToImportUserRequest()
+        internal ImportUserRecordArgsRequest ToImportUserRequest()
         {
-            return new ImportUserRequest(this);
+            return new ImportUserRecordArgsRequest(this);
         }
 
-        private static string UrlSafeBase64Encode(byte[] bytes)
+        internal sealed class ImportUserRecordArgsRequest
```
hiranya911, line 104: May be just call this class `Request` so the full qualified name becomes `ImportUserRecordArgs.Request`. That's much more readable.

FirebaseAdmin/FirebaseAdmin/Auth/Hash/Scrypt.cs
MathBunny: For visibility after our discussion, Newtonsoft by default will Base64 encode byte arrays as shown [here](https://www.newtonsoft.com/json/help/html/SerializationGuide.htm#PrimitiveTypes)

FirebaseAdmin/FirebaseAdmin.Tests/Auth/ImportUserRecordArgsTest.cs
```diff
@@ -0,0 +1,132 @@
...
+                    ProviderId = "google.com",
+                },
+            };
+
+            var customClaims = new Dictionary<string, object>()
+            {
+                { "admin", true },
+            };
+
+            var userMetadata = new UserMetadata(1, 2, null);
```
hiranya911, line 42: We will need to make this constructor public as part of this PR.

FirebaseAdmin/FirebaseAdmin.Tests/Auth/ImportUserRecordArgsTest.cs
```diff
@@ -0,0 +1,132 @@
...
+            var expected = new Dictionary<string, object>()
+            {
+                { "createdAt", userMetadata.CreationTimestamp },
+                { "customAttributes", JsonConvert.SerializeObject(customClaims) },
+                { "disabled", false },
+                { "displayName", "Example" },
+                { "email", "example@gmail.com" },
+                { "emailVerified", true },
+                { "lastLoginAt", userMetadata.LastSignInTimestamp },
+                { "passwordHash", "c2VjcmV0" },
```
hiranya911, line 69: Can these be calculated in the test case instead of hardcoding them?

FirebaseAdmin/FirebaseAdmin.Tests/Auth/ImportUserRecordArgsTest.cs
```diff
@@ -0,0 +1,132 @@
...
+                { "photoUrl", "http://example.com/photo" },
+                { "providerUserInfo", userProviders },
+                { "localId", "123" },
+            };
+
+            Assert.Equal(
+              JsonConvert.SerializeObject(expected),
+              JsonConvert.SerializeObject(importUserRecordArgs.ToRequest()));
+        }
+
```
hiranya911, line 81: Lets also add a test case where `ImportUserRecordArgs` is initialized with the bare minimal set of properties (I think just Uid).

FirebaseAdmin/FirebaseAdmin.Tests/Auth/UserImportHashTest.cs
```diff
@@ -0,0 +1,347 @@
...
+
+using System;
+using System.Collections.Generic;
+using Xunit;
+
+namespace FirebaseAdmin.Auth.Hash.Tests
+{
+    public class UserImportHashTest
+    {
+        private static byte[] signerKey = System.Text.Encoding.UTF8.GetBytes("key%20");
```
hiranya911, line 23: readonly

FirebaseAdmin/FirebaseAdmin.Tests/Auth/ImportUserRecordArgsTest.cs
```diff
@@ -0,0 +1,132 @@
...
+using System.Text;
+using Newtonsoft.Json;
+using Xunit;
+
+namespace FirebaseAdmin.Auth.Tests
+{
+    public class ImportUserRecordArgsTest
+    {
+        [Fact]
+        public void TestImportUserRecordArgsSerializationBasic()
```
hiranya911, line 26: We don't use the `Test*` prefix for .NET test case methods. This can be called `SerializationBasic()`.

FirebaseAdmin/FirebaseAdmin.Tests/Auth/ImportUserRecordArgsTest.cs
```diff
@@ -0,0 +1,132 @@
...
+                { "localId", "123" },
+            };
+
+            Assert.Equal(
+              JsonConvert.SerializeObject(expected),
+              JsonConvert.SerializeObject(importUserRecordArgs.ToRequest()));
+        }
+
+        [Fact]
+        public void TestImportUserRecordArgsMissingUid()
```
hiranya911, line 83: `MissingUid()`

Update the other test case names accordingly.

FirebaseAdmin/FirebaseAdmin.Tests/Auth/UserImportHashTest.cs
```diff
@@ -0,0 +1,347 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using System.Collections.Generic;
+using Xunit;
+
+namespace FirebaseAdmin.Auth.Hash.Tests
```
hiranya911, line 19: Should this file be in a new `Hash` subfolder due to its namespace?

FirebaseAdmin/FirebaseAdmin.Tests/Auth/UserImportHashTest.cs
```diff
@@ -0,0 +1,347 @@
...
+{
+    public class UserImportHashTest
+    {
+        private static byte[] signerKey = System.Text.Encoding.UTF8.GetBytes("key%20");
+        private static byte[] saltSeperator = System.Text.Encoding.UTF8.GetBytes("separator");
+
+        [Fact]
+        public void TestBase()
```
hiranya911, line 27: Mind the names. Remove `Test*` prefix.

FirebaseAdmin/FirebaseAdmin.Tests/Auth/UserImportHashTest.cs
```diff
@@ -0,0 +1,347 @@
...
+                { "memoryCost", 13 },
+            };
+
+            Assert.Equal(expectedResult, scryptHash.GetProperties());
+        }
+
+        [Fact]
+        public void TestStandardScryptHash()
+        {
+            UserImportHash hash = new StandardScrypt()
```
hiranya911, line 68: var

FirebaseAdmin/FirebaseAdmin.Tests/Auth/UserImportHashTest.cs
```diff
@@ -0,0 +1,347 @@
...
+        {
+            var scryptHash = new Scrypt()
+            {
+                Rounds = 9,
+                Key = signerKey,
+                SaltSeparator = saltSeperator,
+                MemoryCost = 3,
+            };
+
+            Assert.Throws<ArgumentException>(() =>
```
hiranya911, line 237: Can be a one-liner:
```
Assert.Throws<ArgumentException>(() => scryptHash.GetProperties());
```

FirebaseAdmin/FirebaseAdmin.Tests/Auth/UserImportOptionsTest.cs
```diff
@@ -0,0 +1,47 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using System.Collections.Generic;
+using Xunit;
+
+namespace FirebaseAdmin.Auth.Hash.Tests
```
hiranya911, line 19: This can probably go in the `FirebaseAdmin.Auth.Tests` namespace.

FirebaseAdmin/FirebaseAdmin.Tests/Auth/UserProviderTest.cs
```diff
@@ -0,0 +1,83 @@
...
+
+using System;
+using Xunit;
+
+namespace FirebaseAdmin.Auth.Tests
+{
+    public class UserProviderTest
+    {
+        [Fact]
+        public void NoUid()
```
hiranya911, line 23: Right 👍

FirebaseAdmin/FirebaseAdmin/Auth/UserProvider.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        /// <summary>
+        /// Gets or sets the user's unique ID assigned by the identity provider. This field is required.
+        /// </summary>
+        [JsonProperty("rawId")]
+        public string Uid
+        {
+            get
+            {
+                if (string.IsNullOrEmpty(this.uid))
```
hiranya911, line 37: Properties with built-in validation is always kindda awkward, and should be avoided whenever possible. Take this scenario for example:

```
var prov = new UserProvider();
if (prov.Uid == null)
{
  prov.Uid = someValue;
}
```

Most developers would expect the above to work, but it will throw because of our validation.

I suggest we introduce an internal nested class (similar to `UserImportRecordArgs.Request`) to handle the serialization. We can do all the validation during the type conversion.

FirebaseAdmin/FirebaseAdmin.IntegrationTests/FirebaseAuthTest.cs
```diff
@@ -510,6 +510,115 @@ public async Task ListUsers()
             }
         }
 
+        [Fact]
+        public async Task ImportUsersBasic()
```
hiranya911, line 514: ImportUsers

FirebaseAdmin/FirebaseAdmin.IntegrationTests/FirebaseAuthTest.cs
```diff
@@ -510,6 +510,115 @@ public async Task ListUsers()
...
+            var args = new ImportUserRecordArgs()
+            {
+                Uid = randomUser.Uid,
+                Email = randomUser.Email,
+                DisplayName = "Random User",
+                PhotoUrl = "https://example.com/photo.png",
+                EmailVerified = true,
+            };
+
+            var options = new UserImportOptions();
```
hiranya911, line 526: Not used

FirebaseAdmin/FirebaseAdmin.IntegrationTests/FirebaseAuthTest.cs
```diff
@@ -510,6 +510,115 @@ public async Task ListUsers()
...
+            IEnumerable<ImportUserRecordArgs> usersLst = new List<ImportUserRecordArgs>();
+            usersLst = usersLst.Append(args);
+
+            var resp = await FirebaseAuth.DefaultInstance.ImportUsersAsync(usersLst);
+
+            try
+            {
+                Assert.Equal(1, resp.SuccessCount);
```
hiranya911, line 534: Also call `GetUserAsync()` on the UID to ensure the user account has been created.

FirebaseAdmin/FirebaseAdmin.IntegrationTests/FirebaseAuthTest.cs
```diff
@@ -510,6 +510,115 @@ public async Task ListUsers()
...
+                Assert.Equal(0, resp.FailureCount);
+            }
+            finally
+            {
+                await FirebaseAuth.DefaultInstance.DeleteUserAsync(randomUser.Uid);
+            }
+        }
+
+        [Fact]
+        public async Task ImportUsersPasswordNoHash()
```
hiranya911, line 544: This ought to be a unit test.

FirebaseAdmin/FirebaseAdmin.IntegrationTests/FirebaseAuthTest.cs
```diff
@@ -510,6 +510,115 @@ public async Task ListUsers()
...
+
+            var options = new UserImportOptions();
+            IEnumerable<ImportUserRecordArgs> usersLst = new List<ImportUserRecordArgs>();
+            usersLst = usersLst.Append(args);
+            await Assert.ThrowsAsync<ArgumentNullException>(
+                async () => await FirebaseAuth.DefaultInstance.ImportUsersAsync(usersLst, options));
+        }
+
+        [Fact]
+        public async Task ImportUsersPassword()
```
hiranya911, line 563: ImportUsersWithPassword

FirebaseAdmin/FirebaseAdmin.IntegrationTests/FirebaseAuthTest.cs
```diff
@@ -510,6 +510,115 @@ public async Task ListUsers()
...
+                PasswordHash = Encoding.ASCII.GetBytes("def"),
+                CustomClaims = new Dictionary<string, object>()
+                {
+                    { "admin", true },
+                },
+                UserProviders = new List<UserProvider>
+                {
+                    new UserProvider()
+                    {
+                        Uid = "google.uid",
```
hiranya911, line 583: Use `randomUser.Uid` and `randomUser.Email` to be on the safe side.

FirebaseAdmin/FirebaseAdmin.IntegrationTests/FirebaseAuthTest.cs
```diff
@@ -510,6 +510,115 @@ public async Task ListUsers()
...
+                        ProviderId = "facebook.com",
+                    },
+                },
+            };
+
+            var options = new UserImportOptions()
+            {
+                Hash = new HmacSha256()
+                {
+                    Key = System.Text.Encoding.UTF8.GetBytes("secretKey"),
```
hiranya911, line 604: Drop `System.Text.`

FirebaseAdmin/FirebaseAdmin.IntegrationTests/FirebaseAuthTest.cs
hiranya911, line 563: I don't think this test case is exactly correct (it's importing a user with incorrect password hashes). Ideally we need something similar to what's in Java. That is import users with a valid Scrypt password hash, and demonstrate that we can login as that user. But feel free to do that change in a future PR. This is acceptable for now.

FirebaseAdmin/FirebaseAdmin/Auth/ImportUserRecordArgs.cs
```diff
@@ -146,41 +147,41 @@ internal Request(ImportUserRecordArgs args)
                 this.Disabled = args.Disabled;
             }
 
-            [JsonProperty("createdAt")]
+            [JsonProperty("createdAt", NullValueHandling = NullValueHandling.Ignore)]
```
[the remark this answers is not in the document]
MathBunny: Yep, that worked!

FirebaseAdmin/FirebaseAdmin.IntegrationTests/FirebaseAuthTest.cs
MathBunny: I will do a separate PR follow this

FirebaseAdmin/FirebaseAdmin.IntegrationTests/FirebaseAuthTest.cs
MathBunny: Should I use `randomUser.Uid` and `randomUser.Email` for the second `UserProvider` as well? Or is it ok to keep it as is? I'm not sure what the difference is

hiranya911: Yeah use random values for all.

FirebaseAdmin/FirebaseAdmin.Tests/Auth/FirebaseAuthTest.cs
```diff
@@ -160,6 +160,26 @@ public async Task SetCustomUserClaimsNoProjectId()
...
+                PasswordSalt = Encoding.ASCII.GetBytes("abc"),
+                PasswordHash = Encoding.ASCII.GetBytes("def"),
+            };
+
+            var options = new UserImportOptions();
+            var usersLst = new List<ImportUserRecordArgs>()
+            {
+                args,
+            };
+            await Assert.ThrowsAsync<NullReferenceException>(
```
hiranya911, line 179: Nit: May be call the overload that doesn't accept `options`.

hiranya911:
@kevinthecheung please review the API docs in all public classes.

PR #206, opened 2020-07-06
fix(auth): Refactored the auth provider (OIDC) management

hiranya911:
There are several readability and maintainability limitations in the current implementation that I'd like to address before implementing the SAML provider config management support.

Issues in the current implementation:
1. Logic for argument validation is split and duplicated across several classes. For example both `ProviderConfigManager` and `OidcProviderConfigArgs` contain logic for validating provider IDs.
2. List configs functionality requires a bunch of parameterized classes in the top-level of the namespace (`ListProviderConfigsRequest<T>`, `ListProviderConfigsPageManager<T>` etc).
3. Some of the parameterized methods in `ProviderConfigManager` have hardcoded logic specific to `AuthProviderConfig` types:

```
// From ProviderConfigManager.CreateProviderConfigAsync<T>():

var query = new Dictionary<string, object>()	           
{	
     { "oauthIdpConfigId", args.ValidateProviderId() },	 // <-- OIDC specific parameter name
};
```

In order to address the above issues, and improve the design I'm implementing the following modifications:

1. A new `ProviderConfigClient<T>` abstraction is introduced. Implementations of this class handle all aspects of argument validation and RPCs related to a specific type of provider configurations.
2. `ProviderConfigManager` now becomes a simple facade, which delegates invocations to the appropriate implementation of `ProviderConfigClient<T>`.
3. All `List*` classes are now nested in the `ProviderConfigClient<T>` hierarchy. This gets them off the top-level of the namespace. Also by having them nested in a generic class make the inheritance work more naturally since each of these classes themselves are generic.
4. A new `ApiClient` abstraction is introduced which handles the low-level stateful HTTP communication.


FirebaseAdmin/FirebaseAdmin/Auth/Providers/ProviderConfigClient.cs
```diff
@@ -0,0 +1,246 @@
...
+            }
+
+            public abstract Task<AuthProviderConfigs<T>> ExecuteAsync(
+                CancellationToken cancellationToken);
+
+            internal void SetPageSize(int? pageSize)
+            {
+                if (pageSize > MaxListResults)
+                {
+                    throw new ArgumentException("Page size must not exceed 100.");
```
MathBunny, line 181: nit: use string interpolation for `MaxListResults`

hiranya911: Done

PR #211, opened 2020-07-15
feat(auth): Added support for updating SAML provider configs

hiranya911:
Implemented `UpdateProviderConfigAsync()` support for `SamlProviderConfigArgs`. There's some room for code de-duplication here, but I'd rather do that refactor in a separate PR.


FirebaseAdmin/FirebaseAdmin.Tests/Auth/Providers/SamlProviderConfigTest.cs
```diff
@@ -404,5 +526,119 @@ public IEnumerator<object[]> GetEnumerator()
...
+                    "X509 certificates must not be empty.",
+                };
+                yield return new object[]
+                {
+                    new SamlProviderConfigArgs()
+                    {
+                        ProviderId = "saml.provider",
+                        X509Certificates = new List<string>() { string.Empty },
+                    },
+                    "X509 certificates must not contain null or empty values.",
```
MathBunny, line 610: Is there a unit test for the null value check?

hiranya911: Done

PR #219, opened 2020-07-24
chore: Moving GetUsersTest to a top-level class

hiranya911:
Auth integration tests are starting to clutter up, so I'm introducing a new `FirebaseAdmin.IntegrationTests.Auth` namespace. 

* `OidcProviderConfigTest` and `SamlProviderConfigTest` have been moved to the new namespace with no modifications.
* `GetUsersTest` which is currently nested in `FirebaseAuthTest` has been promoted to a top-level class, and moved to the new namespace.
* This class also had a TODO to import a test user account, and test for `ProviderIdentifier`, which I've implemented as part of this PR.
* Added a `TemporaryUserBuilder` class for cleanly creating and deleting test user accounts.

In a future PR I will also move `FirebaseAuthTest` into the new namespace, and change it to use the new `TemporaryBuilderClass`.


FirebaseAdmin/FirebaseAdmin.IntegrationTests/Auth/TemporaryUserBuilder.cs
```diff
@@ -0,0 +1,89 @@
...
+        {
+            return await this.CreateUserAsync(RandomUserRecordArgs());
+        }
+
+        public async Task<UserRecord> CreateUserAsync(UserRecordArgs args)
+        {
+            // Make sure we never create more than 1000 users in a single instance.
+            // This allows us to delete all user accounts with a single call to DeleteUsers().
+            // Should not ever occur in practice.
+            if (this.userIds.Count() == 1000)
```
rsgowman, line 67: Is concurrency a concern? I don't think the tests run in parallel, so I would guess no. But if that's wrong, then consider a mutex.

Somewhat related, consider >= 1000 instead of ==, just to be a little extra paranoid.

hiranya911: I thought about the concurrency issue. Tests in the same class are never run in parallel, so this won't be a problem unless we reuse an instance of this class across classes. But I'm inclined to just call this class *not thread safe* and ask the caller synchronize if necessary. I've updated the class summary to that effect.

Changed the userId count check as indicated.

PR #223, opened 2020-07-31
feat(auth): Added CreateTenantAsync() and UpdateTenantAsync() APIs

hiranya911:
* Converted the `Tenant.Args` class to the public `TenantArgs` class.
* Implemented create and update methods for tenants.
* Moved some of the shared http logic to a new `HttpUtils` class.

### DIFF
--- a/FirebaseAdmin/FirebaseAdmin/Auth/Multitenancy/Tenant.cs
+++ b/FirebaseAdmin/FirebaseAdmin/Auth/Multitenancy/Tenant.cs
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 using Google.Apis.Util;
-using Newtonsoft.Json;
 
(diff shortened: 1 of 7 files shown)


FirebaseAdmin/FirebaseAdmin/Auth/Multitenancy/Tenant.cs
```diff
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 using Google.Apis.Util;
```
MathBunny, line 15: Just wondering, is this import being used in this file?

hiranya911: Yes, we use some of its extension methods.

PR #228, opened 2020-08-07
feat(auth): Added support for creating tenant-scoped custom tokens

hiranya911:
* Implemented support for creating tenant-scoped custom tokens.
* Moved custom token creation APIs to the `AbstractFirebaseAuth` parent class so they are available on both `FirebaseAuth` and `TenantAwareFireabseAuth`.
* Cleaned up and refactored the unit tests to check the custom token creation functionality via both public interfaces. See the new `CustomTokenTest` where I use parameterized tests for this purpose.


FirebaseAdmin/FirebaseAdmin.Tests/Auth/CustomTokenTest.cs
```diff
@@ -0,0 +1,145 @@
...
+            {
+                Credential = GoogleCredential.FromAccessToken("test-token"),
+                ProjectId = "project1",
+            };
+            var auth = config.CreateAuth(options);
+
+            var ex = await Assert.ThrowsAsync<InvalidOperationException>(
+                () => auth.CreateCustomTokenAsync("user1"));
+
+            var errorMessage = "Failed to determine service account ID. Make sure to initialize the SDK "
```
MathBunny, line 84: Just a thought, this exact string appears in three other places in the codebase already, maybe if it were a const string the code would be cleaner.

FirebaseAdmin/FirebaseAdmin/Auth/Multitenancy/TenantManager.cs
```diff
@@ -336,8 +337,9 @@ void IDisposable.Dispose()
 
         internal static TenantManager Create(FirebaseApp app)
         {
-            var args = new Args
+            var args = new Args()
```
MathBunny, line 340: Just wondering why was this required?

hiranya911: It's just a cleaner and shorter syntax. We ought to use it whenever possible. Not a must though.

FirebaseAdmin/FirebaseAdmin.Tests/Auth/CustomTokenTest.cs
hiranya911: I believe it's in `IAMSigner` and a couple of test cases. I'd prefer to have the error message repeated in the tests to make sure developers see the correct error message at runtime (it actually contains some useful information for debugging). But if it starts appearing in more places, we can look into introducing a constant of some sorts.

PR #231, opened 2020-08-13
chore: FirebaseAuth unit test refactor

hiranya911:
We have quite a bit of code duplication in our current unit tests, and a lot of unnecessary abstraction. I'm cleaning this up a bit before proceeding to the next stages of the multi-tenancy implementation.

* Moved the test `IPublicKeySource` and `ISigner` to a new `JwtTestUtils` class for easier sharing across test classes.
* Moved the logic in the existing `TestConfig` classes to 2 new classes: `AuthBuilder` and `MockTokenBuilder`. These new classes can be composed in test classes as needed.
* `TestConfig` classes are now simpler due to the above composition. We also no longer need to subclass `TestConfig` in each test class.
* Removing a layer of indirection in `FirebaseTokenVerifierArgs` by dropping the `ForIdTokens()` method. Similarly we can remove the `ForSessionCookies()` method in the future.


FirebaseAdmin/FirebaseAdmin.Tests/Auth/AuthBuilder.cs
```diff
@@ -0,0 +1,92 @@
...
+
+        public AbstractFirebaseAuth Build(TestOptions options)
+        {
+            if (this.TenantId != null)
+            {
+                var args = TenantAwareFirebaseAuth.Args.CreateDefault(this.TenantId);
+                this.PopulateArgs(args, options);
+                return new TenantAwareFirebaseAuth(args);
+            }
+            else
```
MathBunny, line 50: Opinion: The else isn't needed here

FirebaseAdmin/FirebaseAdmin.Tests/Auth/Jwt/MockTokenBuilder.cs
```diff
@@ -0,0 +1,120 @@
...
+
+        internal string IssuerPrefix { get; set; }
+
+        internal IClock Clock { get; set; }
+
+        internal ISigner Signer { get; set; }
+
+        public async Task<string> CreateTokenAsync(
+            IDictionary<string, object> headerOverrides = null,
+            IDictionary<string, object> payloadOverrides = null)
```
lahirumaramba, line 38: Just curious. Should the default of these collections be an empty collection instead of using null?
If they are empty we can avoid the null checks below.

lahirumaramba: I just realized that the default values have to be constants at compile time... unless we use `headerOverrides = headerOverrides ?? new Dictionary <string, object>();`. Please ignore my first comment :)

FirebaseAdmin/FirebaseAdmin.Tests/Auth/AuthBuilder.cs
hiranya911: Wasn't my first choice either. But C# scoping rules do not allow using the same local variable name (args) without the else block. I've renamed the first instance to tenantArgs, and removed the else block.

PR #263, opened 2020-10-16
feat(fcm): Support for sending up to 500 messages in a batch

hiranya911:
RELEASE NOTE: `MulticastMessage` and `SendAllAsync()` APIs now support sending up to 500 messages in a batch.


lahirumaramba (approved):
LGTM! Thanks!

PR #268, opened 2020-10-26
fix(fcm): Setting ApplicationName property for the batch client

### LINKED ISSUE #259: FirebaseMessagingClient does not set Initializer.ApplicationName, causing a warning from BaseClientService
My application uses Firebase Cloud Messaging. When it is started, it logs this warning to a Google.Apis.Logging.ILogger that it registers with Google.ApplicationContext.RegisterLogger:

> Google.Apis.Services.BaseClientService Warning: 0 : Application name is not set. Please set Initializer.ApplicationName property

It seems that the FirebaseMessagingClient constructor calls the FCMClientService constructor with a BaseClientService.Initializer instance whose ApplicationName property is not set: <https://github.com/firebase/firebase-admin-dotnet/blob/e0473fae182d2fbe851839684379298a56f95445/FirebaseAdmin/FirebaseAdmin/Messaging/FirebaseMessagingClient.cs#L67-L71>

That then causes the warning from the BaseClientService constructor: <https://github.com/googleapis/google-api-dotnet-client/blob/703f7d6a34dcbf06e09ff7e36e59d747940f0c84/Src/Support/Google.Apis/Services/BaseClientService.cs#L164-L169>

I don't see any way for my application to set Initializer.ApplicationName and thus avoid the issue that causes the warning. I suppose I could change the Google.Apis.Logging.ILogger implementation in my application to recognize and suppress this specific warning.

My application hit this issue with the following set of NuGet packages. These versions are a bit out of date but the relevant parts of the source code don't seem to have been changed since then.

* FirebaseAdmin/1.9.0
* Google.Api.Gax/2.7.0
* Google.Api.Gax.Rest/2.7.0
* Google.Apis/1.41.1
* Google.Apis.Auth/1.41.1
* Google.Apis.Core/1.41.1

hiranya911:
RELEASE NOTE: Setting the `ApplicationName` property for the batch request client, which resolves a warning log message seen in previous versions of the SDK.

Resolves #259

### DIFF
--- a/FirebaseAdmin/FirebaseAdmin.Tests/Messaging/FirebaseMessagingClientTest.cs
+++ b/FirebaseAdmin/FirebaseAdmin.Tests/Messaging/FirebaseMessagingClientTest.cs
@@ -194,6 +194,8 @@ Vary: Referer
             Assert.Equal("projects/fir-adminintegrationtests/messages/5903525881088369386", response.Responses[1].MessageId);
             Assert.Equal(1, handler.Calls);
 
+            var userAgent = handler.LastRequestHeaders.UserAgent.First();
+            Assert.Equal("fire-admin-dotnet", userAgent.Product.Name);
             Assert.Equal(2, this.CountLinesWithPrefix(handler.LastRequestBody, VersionHeader));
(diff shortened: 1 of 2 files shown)


lahirumaramba (approved):
Thanks! LGTM!

PR #280, opened 2020-12-31
Add DateTimeOffset property for simpler usage

Thaina:


hiranya911:
This requires tests.

Also since this is adding to a public API surface, it needs to go through our internal API review process, which can take some time.

Issue #71, opened 2019-06-10
FR: Support Cloud Storage


takuya1981:
firebase-admin-dotnet does not support Cloud Storage yet. 
I'd like to contribute for it.

The simple specifications are as follows:
* Need new class to handle Cloud Storage. Its name is FirebaseAdmin.Cloud.FirebaseStorage as an example.
* FirebaseStorage instance is managed in FirebaseApp class.
* FirebaseStorage instance has Google.Cloud.Storage.V1.StorageClient instance.
* FirebaseStorage instance will return the StorageClient instance associated with the Firebase app.
* Now the Bucket class is returned in other Java/Python SDK. But the Bucket class in .NET is almost data class. It's not neccesary to be returned(It's my opinion).
* When calling FirebaseApp.Delete(), StorageClient.Dispose() is called.

I would like to open a pull request about the above if there are no problems.

hiranya911:
Seems like a good addition. Except I wouldn't call the class `FirebaseStorage`. Perhaps `StorageClientHelper` or just `GoogleCloudStorage`.

```
using FirebaseAdmin.Cloud;
using Google.Cloud.Storage.V1;

StorageClient client = StorageClientHelper.DefaultInstance.GetClient();
```

takuya1981:
Name is StorageClientHelper, I understand.

StorageClient support custom encryption key like this.
```
public static StorageClient Create(GoogleCredential credential = null, EncryptionKey encryptionKey = null);
```
This class should support this function like this.
```
public static StorageClient GetStorageClient(EncryptionKey encryptionKey = null){}
public static StorageClient GetStorageClient(FirebaseApp app, EncryptionKey encryptionKey = null)
```
What do you think?

hiranya911:
Hi @takuya1981. Thanks again for thinking about this. I'm not convinced that we should expose custom encryption keys at this stage. We haven't exposed it in other languages. Therefore let's start with just `GetStorageClient()` for now. Note that `GetStorageClient(FirebaseApp)` won't be necessary if we attach `CloudStorageHelper` instance to a `FirebaseApp`. So we are looking at something like:

```
public class CloudStorageHelper
   public static CloudStorageHelper DefaultInstance;
   public static CloudStorageHelper GetCloudStorageHelper(FirebaseApp app);
   public StorageClient GetStorageClient();
```

We can support encryption keys in the future if we want to, but that should be supported across all languages if we do that.

hiranya911:
[...]

PR #286, opened 2021-02-24
Implementation of Cloud Storage function.

kiliczsh:
Original PR: https://github.com/firebase/firebase-admin-dotnet/pull/72 #71 
Configuration keys are missing on the actions machines so that had to handle exceptions.
Google was not allowing me to submit other users commits so that had to copy all, sorry @takuya1981. 
@googlebot I consent.

All credits to @takuya1981 .


hiranya911:
Thanks @kiliczsh for the PR (complete with integration tests). Storage support in .NET has always been a bit of a conundrum. In all other languages of the Admin SDK, we expose the "bucket" as the primary abstraction. For example in Node.js:

```js
const admin = require('firebase-admin');
admin.initializeApp({
  storageBucket: 'example.bucket.name',
});

const bucket = admin.storage().bucket();
```

But as you already know, in the .NET Storage client, that abstraction does not exist. This results in a very different developer experience. I'd like to run this by our internal API review team, and see what they think about it. Since we have no way of passing a bucket name to the Storage client via Firebase app options, I'm not certain if this integration adds a lot of value.

PR #291, opened 2021-04-01
feat(auth): Supporting VerifyIdToken() API with emulator

hiranya911:
Added emulator support for `VerifyIdTokenAsync()` method. In the emulator mode:

* Algorithm and KID header validations are skipped
* Signature verification is skipped


lahirumaramba (approved):
LGTM!

PR #293, opened 2021-04-06
feat(auth): Emulator support for CreateCustomTokenAsync() API

hiranya911:
Implemented support for creating custom tokens for the emulator. In the emulator mode:

* The `alg` header is set to `none`.
* Signature is set to empty byte sequence.

Bulk of the changes are just refactoring the existing tests. I've updated `CustomTokenTest` to be more inline with other test classes like `IdTokenVerificationTest`, where we test behavior via parameterization, as opposed to creating `FirebaseApp` instances with different configurations.

### DIFF
--- a/FirebaseAdmin/FirebaseAdmin/Auth/Jwt/EmulatorSigner.cs
+++ b/FirebaseAdmin/FirebaseAdmin/Auth/Jwt/EmulatorSigner.cs
@@ -1,0 +1,42 @@
+// Copyright 2021`, Google Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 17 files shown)


FirebaseAdmin/FirebaseAdmin/Auth/Jwt/EmulatorSigner.cs
```diff
@@ -0,0 +1,42 @@
+// Copyright 2021`, Google Inc. All rights reserved.
```
lahirumaramba, line 1: nit: extra `

lahirumaramba (approved):
LGTM!
Is there a `kid` in the header for emulator tokens?

FirebaseAdmin/FirebaseAdmin/Auth/Jwt/EmulatorSigner.cs
hiranya911: Thanks. Done.

PR #298, opened 2021-05-03
fix(auth): adds missing EMAIL_NOT_FOUND error code

bojeil-google:
Catch EMAIL_NOT_FOUND when /accounts:sendOobCode is called for password reset on a user that does not exist.

### DIFF
--- a/FirebaseAdmin/FirebaseAdmin/Auth/AuthErrorCode.cs
+++ b/FirebaseAdmin/FirebaseAdmin/Auth/AuthErrorCode.cs
@@ -30,13 +30,6 @@ namespace FirebaseAdmin.Auth
         EmailAlreadyExists,
 
         /// <summary>
-        /// No user record found for the given email, typically raised when
-        /// generating a password reset link using an email for a user that
-        /// is not already registered.
-        /// </summary>
-        EmailNotFound,
-
-        /// <summary>
         /// The specified ID token is expired.
         /// </summary>
         ExpiredIdToken,
@@ -105,5 +98,12 @@ namespace FirebaseAdmin.Auth
         /// Tenant ID in a token does not match.
         /// </summary>
         TenantIdMismatch,
+
+        /// <summary>
+        /// No user record found for the given email, typically raised when
+        /// generating a password reset link using an email for a user that
+        /// is not already registered.
+        /// </summary>
+        EmailNotFound,
     }
 }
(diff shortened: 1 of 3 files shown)


FirebaseAdmin/FirebaseAdmin/Auth/AuthErrorCode.cs
```diff
@@ -29,6 +29,13 @@ public enum AuthErrorCode
         /// </summary>
         EmailAlreadyExists,
 
+        /// <summary>
```
hiranya911, line 32: Add to the end of this enum. C# assigns an integer to each enum value depending on their position, and also lets developers access that integer value:

```cs
var code = AuthErrorCode.UserNotFound;
Console.WriteLine((int)code);
```

It is a breaking change to modify the int values (kind of like the field position in protos)

FirebaseAdmin/FirebaseAdmin/Auth/AuthErrorHandler.cs
```diff
@@ -63,6 +63,13 @@ internal sealed class AuthErrorHandler
                         AuthErrorCode.EmailAlreadyExists,
                         "The user with the provided email already exists")
                 },
+                {
+                    "EMAIL_NOT_FOUND",
```
hiranya911, line 67: Also update the test case at: https://github.com/firebase/firebase-admin-dotnet/blob/e71b52a310b2e7bab461a51f4f3212cc7afa76d6/FirebaseAdmin/FirebaseAdmin.Tests/Auth/AuthErrorHandlerTest.cs#L27

hiranya911 (changes requested):
Couple of changes suggested.

FirebaseAdmin/FirebaseAdmin/Auth/AuthErrorCode.cs
bojeil-google: Thanks for catching and explaining. I moved it to the end of the enum.

FirebaseAdmin/FirebaseAdmin/Auth/AuthErrorHandler.cs
bojeil-google: Done

hiranya911 (approved):
LGTM 👍

egilmorez (approved):

PR #302, opened 2021-08-02
Android notification EventTime property

patrick11994:
Added the `EventTime` property to `AndroidNotification` so it can be set. 

This is basically the equivalent of this code in Java:
```java
androidBuilder.setEventTimeInMillis(SomeValue);
```

I also took the [documentation of the Java-side](https://github.com/firebase/firebase-admin-java/blob/master/src/main/java/com/google/firebase/messaging/AndroidNotification.java#L448-L458) and changed it a bit to comply with the code style of this project.

Tested on my local machine with sending notifications and it works like a charm 😄


lahirumaramba:
Thank you @patrick11994 for your contribution! We really appreciate it. However, this feature is covered in #317

patrick11994:
Thanks, it is OK. I hope that this other PR gets merged soon.

PR #343, opened 2023-03-29
feat(fcm): Implement `SendEachAsync` and `SendEachForMulticastAsync`

Doris-Ge:
`SendEachAsync` vs `SendAllAsync`
1. `SendEachAsync` sends one HTTP request to V1 Send endpoint for each message in the list. `SendAllAsync` sends only one HTTP request to V1 Batch Send endpoint to send all messages in the list.
2. `SendEachAsync` calls `Task.WhenAll` to wait for all `httpClient.SendAndDeserializeAsync` calls to complete and construct a `BatchResponse` with all `SendResponse`s. An `httpClient.SendAndDeserializeAsync` call to V1 Send endpoint either completes with a success or throws an exception. So if an exception is thrown out, the exception will be caught in `SendEachAsync` and turned into a `SendResponse` with an exception. Therefore, unlike `SendAllAsync`, `SendEachAsync` does not always throw an exception for a total failure. It can also return a `BatchResponse` with only exceptions in it.

`SendEachForMulticastAsync` calls `SendEachAsync` under the hood.

### DIFF
--- a/FirebaseAdmin/FirebaseAdmin/Messaging/FirebaseMessagingClient.cs
+++ b/FirebaseAdmin/FirebaseAdmin/Messaging/FirebaseMessagingClient.cs
@@ -127,6 +127,49 @@ namespace FirebaseAdmin.Messaging
         }
 
         /// <summary>
+        /// Sends each message in a single batch.
+        /// </summary>
+        /// <param name="messages">The messages to be sent. Must not be null.</param>
+        /// <param name="dryRun">A boolean indicating whether to perform a dry run (validation
+        /// only) of the send. If set to true, the messages will be sent to the FCM backend service,
+        /// but it will not be delivered to any actual recipients.</param>
+        /// <param name="cancellationToken">A cancellation token to monitor the asynchronous
+        /// operation.</param>
+        /// <returns>A task that completes with a <see cref="BatchResponse"/>, giving details about
+        /// the batch operation.</returns>
+        public async Task<BatchResponse> SendEachAsync(
+            IEnumerable<Message> messages,
+            bool dryRun = false,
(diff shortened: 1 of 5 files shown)


FirebaseAdmin/FirebaseAdmin/Messaging/FirebaseMessagingClient.cs
```diff
@@ -251,6 +294,31 @@ private static void AddCommonHeaders(HttpRequestMessage request)
...
+            {
+                return SendResponse.FromException(e);
+            }
+            catch (HttpRequestException e)
+            {
+                return SendResponse.FromException(MessagingErrorHandler.Instance.HandleHttpRequestException(e));
+            }
+            catch (Exception e)
+            {
+                var exception = new FirebaseMessagingException(ErrorCode.Unknown, $"{e}");
```
Doris-Ge, line 317: Is it ok to put the exception into the `SendResponse`?

lahirumaramba: Seems like that's how we have handled exceptions in batch send. Add the responses in `var responses = new List<SendResponse>();` with `SendResponse.FromException(...` for failed once.

lahirumaramba (approved):
LGTM! Thank you!

FirebaseAdmin/FirebaseAdmin/Messaging/FirebaseMessagingClient.cs
Doris-Ge: Sorry for the late response! My question was about whether it's ok to put the content of the `Exception e` into the `SendResponse` via `$"{e}"`. Sorry for being unclear!

PR #348, opened 2023-06-08
feat(fcm): Add `SendEachAsync` and `SendEachForMulticastAsync`  for FCM batch send

Doris-Ge:
* Deprecate `SendAllAsync`() and `SendMulticastAsync` APIs.
* Add `SendEachAsync` and `SendEachForMulticastAsync` APIs

`SendEachAsync` vs `SendAllAsync`
1. `SendEachAsync` sends one HTTP request to V1 Send endpoint for each message in the list. `SendAllAsync` sends only one HTTP request to V1 Batch Send endpoint to send all messages in the list.
2. `SendEachAsync` calls `Task.WhenAll` to wait for all `httpClient.SendAndDeserializeAsync` calls to complete and construct a `BatchResponse` with all `SendResponse`s. An `httpClient.SendAndDeserializeAsync` call to V1 Send endpoint either completes with a success or throws an exception. So if an exception is thrown out, the exception will be caught in `SendEachAsync` and turned into a `SendResponse` with an exception. Therefore, unlike `SendAllAsync`, `SendEachAsync` does not always throw an exception for a total failure. It can also return a `BatchResponse` with only exceptions in it.

`SendEachForMulticastAsync` calls `SendEachAsync` under the hood.

RELEASE NOTE: `SendAllAsync()` and `SendMulticastAsync()` APIs are now deprecated. Use `SendEachAsync()` and `SendEachForMulticastAsync()` APIs instead.

### DIFF
--- a/FirebaseAdmin/FirebaseAdmin/Messaging/FirebaseMessaging.cs
+++ b/FirebaseAdmin/FirebaseAdmin/Messaging/FirebaseMessaging.cs
@@ -173,12 +173,13 @@ namespace FirebaseAdmin.Messaging
 
         /// <summary>
         /// Sends each message in the given list via Firebase Cloud Messaging. Unlike
-        /// <see cref="SendAllAsync(IEnumerable{Message})"/>, this method makes a single HTTP call
+        /// <see cref="SendAllAsync(IEnumerable{Message})"/>, this method makes an HTTP call
         /// for each message in the given list.
         /// </summary>
         /// <exception cref="FirebaseMessagingException">If an error occurs while sending the
         /// messages.</exception>
-        /// <param name="messages">Up to 500 messages to send in the batch. Cannot be null.</param>
+        /// <param name="messages">Up to 500 messages to send in the batch. Cannot be null or
+        /// empty.</param>
         /// <returns>A <see cref="BatchResponse"/> containing details of the batch operation's
         /// outcome.</returns>
         public async Task<BatchResponse> SendEachAsync(IEnumerable<Message> messages)
@@ -189,12 +190,13 @@ namespace FirebaseAdmin.Messaging
 
         /// <summary>
         /// Sends each message in the given list via Firebase Cloud Messaging. Unlike
-        /// <see cref="SendAllAsync(IEnumerable{Message}, CancellationToken)"/>, this method makes a
-        /// single HTTP call for each message in the given list.
+        /// <see cref="SendAllAsync(IEnumerable{Message}, CancellationToken)"/>, this method makes
+        /// an HTTP call for each message in the given list.
         /// </summary>
         /// <exception cref="FirebaseMessagingException">If an error occurs while sending the
         /// messages.</exception>
-        /// <param name="messages">Up to 500 messages to send in the batch. Cannot be null.</param>
+        /// <param name="messages">Up to 500 messages to send in the batch. Cannot be null or
+        /// empty.</param>
         /// <param name="cancellationToken">A cancellation token to monitor the asynchronous
         /// operation.</param>
         /// <returns>A <see cref="BatchResponse"/> containing details of the batch operation's
@@ -207,12 +209,13 @@ namespace FirebaseAdmin.Messaging
 
         /// <summary>
         /// Sends each message in the given list via Firebase Cloud Messaging. Unlike
-        /// <see cref="SendAllAsync(IEnumerable{Message}, bool)"/>, this method makes a
-        /// single HTTP call for each message in the given list.
+        /// <see cref="SendAllAsync(IEnumerable{Message}, bool)"/>, this method makes an HTTP
+        /// call for each message in the given list.
         /// </summary>
         /// <exception cref="FirebaseMessagingException">If an error occurs while sending the
         /// messages.</exception>
-        /// <param name="messages">Up to 500 messages to send in the batch. Cannot be null.</param>
+        /// <param name="messages">Up to 500 messages to send in the batch. Cannot be null or
+        /// empty.</param>
         /// <param name="dryRun">A boolean indicating whether to perform a dry run (validation
         /// only) of the send. If set to true, the message will be sent to the FCM backend service,
         /// but it will not be delivered to any actual recipients.</param>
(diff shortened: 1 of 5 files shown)


FirebaseAdmin/FirebaseAdmin/Messaging/FirebaseMessaging.cs
```diff
@@ -171,16 +171,175 @@ public async Task<string> SendAsync(Message message, bool dryRun)
                 message, dryRun, cancellationToken).ConfigureAwait(false);
         }
 
+        /// <summary>
+        /// Sends each message in the given list via Firebase Cloud Messaging. Unlike
+        /// <see cref="SendAllAsync(IEnumerable{Message})"/>, this method makes a single HTTP call
```
kevinthecheung, line 176:
```suggestion
        /// <see cref="SendAllAsync(IEnumerable{Message})"/>, this method makes an HTTP call
```

FirebaseAdmin/FirebaseAdmin/Messaging/FirebaseMessaging.cs
```diff
@@ -171,16 +171,175 @@ public async Task<string> SendAsync(Message message, bool dryRun)
...
+        /// for each message in the given list.
+        /// </summary>
+        /// <exception cref="FirebaseMessagingException">If an error occurs while sending the
+        /// messages.</exception>
+        /// <param name="messages">Up to 500 messages to send in the batch. Cannot be null.</param>
```
kevinthecheung, line 181: empty ok?

FirebaseAdmin/FirebaseAdmin/Messaging/FirebaseMessaging.cs
```diff
@@ -171,16 +171,175 @@ public async Task<string> SendAsync(Message message, bool dryRun)
...
+        public async Task<BatchResponse> SendEachAsync(IEnumerable<Message> messages)
+        {
+            return await this.SendEachAsync(messages, false)
+                .ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Sends each message in the given list via Firebase Cloud Messaging. Unlike
+        /// <see cref="SendAllAsync(IEnumerable{Message}, CancellationToken)"/>, this method makes a
+        /// single HTTP call for each message in the given list.
```
kevinthecheung, line 193:
```suggestion
        /// <see cref="SendAllAsync(IEnumerable{Message}, CancellationToken)"/>, this method makes an
        /// HTTP call for each message in the given list.
```

FirebaseAdmin/FirebaseAdmin/Messaging/FirebaseMessaging.cs
```diff
@@ -171,16 +171,175 @@ public async Task<string> SendAsync(Message message, bool dryRun)
...
+        public async Task<BatchResponse> SendEachAsync(IEnumerable<Message> messages, CancellationToken cancellationToken)
+        {
+            return await this.SendEachAsync(messages, false, cancellationToken)
+                .ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Sends each message in the given list via Firebase Cloud Messaging. Unlike
+        /// <see cref="SendAllAsync(IEnumerable{Message}, bool)"/>, this method makes a
+        /// single HTTP call for each message in the given list.
```
kevinthecheung, line 211:
```suggestion
        /// <see cref="SendAllAsync(IEnumerable{Message}, bool)"/>, this method makes an
        /// HTTP call for each message in the given list.
```

FirebaseAdmin/FirebaseAdmin/Messaging/FirebaseMessaging.cs
```diff
@@ -171,16 +171,175 @@ public async Task<string> SendAsync(Message message, bool dryRun)
...
+        public async Task<BatchResponse> SendEachAsync(IEnumerable<Message> messages, bool dryRun)
+        {
+            return await this.SendEachAsync(messages, dryRun, default)
+                .ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Sends each message in the given list via Firebase Cloud Messaging. Unlike
+        /// <see cref="SendAllAsync(IEnumerable{Message}, bool, CancellationToken)"/>, this method
+        /// makes a single HTTP call for each message in the given list.
```
kevinthecheung, line 230:
```suggestion
        /// makes an HTTP call for each message in the given list.
```

kevinthecheung (approved):

FirebaseAdmin/FirebaseAdmin/Messaging/FirebaseMessaging.cs
Doris-Ge: I'm not sure what you mean. Could you clarify?

kevinthecheung: non-null and non-empty? or just non null?

Doris-Ge: I think empty is also not ok. It's the same for `SendAllAsync`, but we didn't explicitly mention this for `SendAllAsync` in the past. Should I change it to "Cannot be null or empty."  for both all `SendAll` APIs and all `SendEach` APIs? Or I can keep `SendAll` API as they are now and only update the comment for `SendEach` APIs?

Doris-Ge: I kept `SendAll` API unchanged and updated to "Cannot be null or empty."  for all `SendEach` APIs.

lahirumaramba (approved):
Thank you!

hankovich:
@Doris-Ge Could you please clarify why sending N requests with 1 notification in each is better rather then sending 1 request with N notifications? For me that fact that I should use `SendEachAsync` instead of `SendAllAsync` sounds counterintuitive

Doris-Ge:
@hankovich `SendAllAsync` calls the legacy Batch Send API [which will be discontinued starting June 20, 2024](https://firebase.google.com/support/faq#fcm-23-deprecation). That's the main reason why it is replaced by `SendEachAsync`. `SendEachAsync`calls the HTTP v1 API which does not support sending 1 request with N notifications.

naveedahmed1:
Do we have similar option available in .net sdk? https://github.com/firebase/firebase-admin-node/issues/2488#issuecomment-2251304450

PR #358, opened 2023-08-09
New snippet with project id defined.

pwujczyk:
In the internal bug b/290344815 we had been discussing how to initialize SDK and we come up with working version which is different than documentation on the page: https://firebase.google.com/docs/admin/setup?hl=pl#initialize_the_sdk_in_non-google_environments.

This is only part of the change. The second part is done in cider.  cl/555243967


lahirumaramba (approved):
Thank you! LGTM. Please sign the CLA.

pwujczyk:
done.

Xiaoshouzi-gh (approved):
Thank you for the new snippet!

Issue #352, opened 2023-06-27
Transitive dependency with high security vulnerabilities


Misiu:
I've run `dotnet list package --vulnerable --include-transitive`
and got this output:
```
Użyto następujących źródeł:
   https://api.nuget.org/v3/index.json

Projekt „Test.Consumer” ma następujące pakiety podatne na zagrożenia
   [net7.0]:
   Pakiet przechodni      Rozpoznane   Ważność   Adres URL porady
   > Newtonsoft.Json      12.0.3       High      https://github.com/advisories/GHSA-5crp-9r3c-p9vr
```

I'm currently using FirebaseAdmin version 2.3.0

Ideally, Newtonsoft.Json should be removed in favor of System.Text.Json (or maybe we could pick the serializer that we want to use).

josbol:
Any updates on this? The dependencies on Google.API.* are pointing to very old versions (2020).

PR #377, opened 2024-02-02
chore: Bump supported .NET versions to `net462` and `net6.0`+

jonathanedey:
- Dropped `net461` support
- Added `net462` and `net6.0`+ support

### DIFF
--- a/.github/scripts/builder.sh
+++ b/.github/scripts/builder.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+set -e
+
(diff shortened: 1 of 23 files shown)


Misiu:
any chance this might fix the Newtonsoft.Json issue?
https://github.com/firebase/firebase-admin-dotnet/issues/352

jonathanedey:
Thanks for bring this up @Misiu. Once these upgrades are finished, that and similar issues should be resolved.

PR #394, opened 2024-04-18
chore: fix dependabot .NET path

jonathanedey:
Should now point to `FirebaseAdmin/FirebaseAdmin.sln` which has references to each of the nested `*.csproj` file.


lahirumaramba (approved):
Thanks!

PR #442, opened 2025-01-02
chore: Add mono install step to GHAs

jonathanedey:
Ubuntu 24.04 requires mono to be installed in order to run tests in net462.


lahirumaramba (approved):
Thanks Jonathan!

PR #498, opened 2026-01-16
chore: Enable NuGet trusted publishing

jonathanedey:
See: https://learn.microsoft.com/en-us/nuget/nuget-org/trusted-publishing

Also removed unused release tweet action.


lahirumaramba (approved):
Thanks! LGTM!

PR #525, opened 2026-06-05
feat(fcm): Enable `fid` and deprecate `token` for Send API

yvonnep165:
This change deprecates `Token` (in `Message`) and `Tokens` (in `MulticastMessage`) as message targets, transitioning support to `Fid` (Firebase Installation ID) and `Fids` as the primary targets instead.

Legacy token tests are wrapped in Method-Level #pragma warning disable CS0618 to prevent compiler warnings and build crash and #pragma wrappers are explicitly maintained for snippet documentation code to allow compiling flawlessly.

### DIFF
--- a/FirebaseAdmin/FirebaseAdmin.Snippets/FirebaseExceptionSnippets.cs
+++ b/FirebaseAdmin/FirebaseAdmin.Snippets/FirebaseExceptionSnippets.cs
@@ -139,7 +139,9 @@ namespace FirebaseAdmin.Snippets
         {
             return new Message()
             {
+#pragma warning disable CS0618
                 Token = deviceToken,
+#pragma warning restore CS0618
                 Notification = new Notification()
                 {
                     Title = "Test notification",
--- a/FirebaseAdmin/FirebaseAdmin.Snippets/FirebaseExceptionSnippets.cs
+++ b/FirebaseAdmin/FirebaseAdmin.Snippets/FirebaseExceptionSnippets.cs
@@ -102,6 +102,40 @@ namespace FirebaseAdmin.Snippets
             // [END platform_error_code]
         }
 
+        internal static async Task PlatformErrorCodeWithFid(string fid)
+        {
+            // [START platform_error_code_fid]
+            var notification = CreateNotificationWithFid(fid);
+            try
+            {
+                await FirebaseMessaging.DefaultInstance.SendAsync(notification);
+            }
+            catch (FirebaseMessagingException ex)
+            {
+                // All exceptions contain a platform-level error code. Applications can inspect
+                // both the platform-level error code and any service-level error codes when
+                // implementing error handling logic.
+                if (ex.MessagingErrorCode == MessagingErrorCode.Unregistered)
+                {
+                    // Service-level error code
+                    Console.WriteLine("App instance has been unregistered");
+                    RemoveFidFromDatabase(fid);
+                }
+                else if (ex.ErrorCode == ErrorCode.Unavailable)
+                {
+                    // Platform-level error code
+                    Console.WriteLine("FCM service is temporarily unavailable");
+                    ScheduleForRetry(notification, TimeSpan.FromHours(1));
+                }
+                else
+                {
+                    Console.WriteLine($"Failed to send notification: {ex.Message}");
+                }
+            }
+
+            // [END platform_error_code_fid]
+        }
+
         internal static async Task HttpResponse(string deviceToken)
         {
             // [START http_response]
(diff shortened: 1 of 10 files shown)


FirebaseAdmin/FirebaseAdmin/Messaging/MulticastMessage.cs
Doris-Ge, line 1: Shall we update the comments for `SendEachForMulticastAsync` to mention the message response order in `BatchResponse` when both `Tokens` and `FIDs` are set in `MulticastMessage`?

FirebaseAdmin/FirebaseAdmin.Tests/Messaging/FirebaseMessagingClientTest.cs
Doris-Ge, line 1: Could we update this test file to include some tests using fid messages?

FirebaseAdmin/FirebaseAdmin/Messaging/MulticastMessage.cs
yvonnep165: Sure! Updated comments for both `SendEachForMulticastAsync` and `SendMulticastAsync`

FirebaseAdmin/FirebaseAdmin.Tests/Messaging/FirebaseMessagingClientTest.cs
yvonnep165: `MessageTest.cs` is already handling the payload serialization/deserialization logic for FIDs. `FirebaseMessagingClientTest.cs` is responsible for testing the HTTP transport layer. It typically uses a generic `Topic` or `Token` payload just as a dummy placeholder to verify that the HTTP request is built correctly. Since we already have full FID coverage in `MessageTest.cs` and the `FirebaseMessagingTest.cs` integration tests, it seems a bit redundant to add some tests using fid messages in `FirebaseMessagingClientTest.cs`. But please let me know if you'd still prefer I add a dummy FID payload test there for completeness!

Doris-Ge: If we prefer not to add new tests, it would also be nice to simply replace the Token payload with the FID payload as a dummy placeholder. Since the Token payload is deprecated, this would also avoid the `#pragma warning`. Additionally, this covers testing the integration with the FID payload.

yvonnep165: Sounds good! I have replaced the Token payload with the FID one. Thanks for the detailed review!

Doris-Ge (approved):

FirebaseAdmin/FirebaseAdmin.Snippets/FirebaseExceptionSnippets.cs
```diff
@@ -139,7 +139,9 @@ private static Message CreateNotification(string deviceToken)
         {
             return new Message()
             {
+#pragma warning disable CS0618
                 Token = deviceToken,
+#pragma warning restore CS0618
                 Notification = new Notification()
```
lahirumaramba, line 210: Can you add extend the code snippet to show the new usage with Fids?

FirebaseAdmin/FirebaseAdmin.Tests/Messaging/FirebaseMessagingClientTest.cs
```diff
@@ -197,8 +197,8 @@ public async Task SendEachAsyncWithError()
                     {
                         return @"{
                                     ""error"": {
-                                        ""status"": ""INVALID_ARGUMENT"",
```
lahirumaramba, line 200: What is the reason for removing the `INVALID_ARGUMENT` error code? I think it is good to verify that a 400 Bad Request with `INVALID_ARGUMENT` error converts to `MessagingErrorCode.Unregistered` (for tokens) unless the recent fix changes that for tokens as well.

lahirumaramba (approved):

FirebaseAdmin/FirebaseAdmin.Snippets/FirebaseExceptionSnippets.cs
yvonnep165: Sure! Code snippets for the new usage with Fids are added.

FirebaseAdmin/FirebaseAdmin.Tests/Messaging/FirebaseMessagingClientTest.cs
yvonnep165: I was doing a complete replacement from Token to Fid in this file but you're right, it's good to verify the bad request for tokens. I have added them back.

FirebaseAdmin/FirebaseAdmin/Messaging/Message.cs
```diff
@@ -116,24 +124,27 @@ private string UnprefixedTopic
...
-                copy.Token, copy.Topic, copy.Condition,
+                copy.Token, copy.Fid, copy.Topic, copy.Condition,
             };
+#pragma warning restore CS0618
             var targets = list.FindAll((target) => !string.IsNullOrEmpty(target));
             if (targets.Count != 1)
             {
                 throw new ArgumentException(
-                    "Exactly one of Token, Topic or Condition is required.");
+                    "Exactly one of Token, Fid, Topic or Condition is required.");
```
OrlandriaH-G:
```suggestion
                    "Exactly one of Token, FID, Topic or Condition is required.");
```

FirebaseAdmin/FirebaseAdmin/Messaging/MulticastMessage.cs
```diff
@@ -58,14 +64,27 @@ public sealed class MulticastMessage
...
 
-            if (tokens == null || tokens.Count > 500)
+            var tokensCount = tokensCopy?.Count ?? 0;
+            var fidsCount = fidsCopy?.Count ?? 0;
+            var totalCount = tokensCount + fidsCount;
+
+            if (totalCount == 0)
             {
-                throw new ArgumentException("Tokens must be non-null and contain at most 500 tokens.");
+                throw new ArgumentException("Tokens and Fids cannot be both null or empty.");
```
OrlandriaH-G:
```suggestion
                throw new ArgumentException("Tokens and FIDs cannot be both null or empty.");
```

FirebaseAdmin/FirebaseAdmin/Messaging/MulticastMessage.cs
```diff
@@ -58,14 +64,27 @@ public sealed class MulticastMessage
...
             }
 
-            var tokensCopy = new List<string>(tokens);
+            if (totalCount > 500)
+            {
+                throw new ArgumentException("Total number of Tokens and Fids must not exceed 500.");
```
OrlandriaH-G:
```suggestion
                throw new ArgumentException("Total number of Tokens and FIDs must not exceed 500.");
```

OrlandriaH-G (approved):

